PR #12, opened 2021-11-04
ci(lint): adding a gha for linting

### LINKED ISSUE #9: github action for linting checks
# Summary

Linting can be a contentious topic during code reviews and can slow down the overall review process, focusing less on the design and implementation details, and more on style conflicts. Adding a static linting check can help alleviate that part of the review process entirely, and make the codebase more consistent.

Kubebuilder already scaffolds a root Makefile that contains a `lint` check, along with other related targets that help ensure maintainability: https://github.com/operator-framework/combo/blob/main/Makefile#L26-L33.

# Acceptance Criteria

Introducing a linting-style github action that runs on every PR. That workflow is responsible for checking whether there are any diffs that are produced when running any of the linting/formatting/etc. targets from the root Makefile. If any of those operations produce a diff, then return a non-zero exit code.

## Example(s)

- Run `go mod tidy` and check whether there are any root go.mod/go.tidy changes that need to be checked in
- Check whether the generated files need to be updated and checked into the branch
- Check whether any of the linting or formatting checks produce a diff (or return a non-zero exit code)

Concrete example from upstream OLM: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/.github/workflows/sanity.yaml#L22.

tylerslaton:
# Summary
Adding a new CI action that runs the golangci-lint tool and fix the issues it output.

# Additions
- New CI action for linting and verifying code
- Fixes to appease the linter

Closes #9

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,21 @@
+name: ci
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    paths-ignore:
+    - '**/*.md'
+jobs:
+  sanity:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v2
+      - name: Install Go 1.17
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.17'
+      - name: Run sanity checks
+        run: |
+          make verify && make lint
--- a/main.go
+++ b/main.go
@@ -1,9 +1,13 @@
 package main
 
 import (
+	"log"
+
 	"github.com/operator-framework/combo/cmd"
 )
 
 func main() {
-	cmd.Execute()
+	if err := cmd.Execute(); err != nil {
+		log.Fatal(err)
+	}
 }
--- a/main.go
+++ b/main.go
@@ -1,13 +1,15 @@
 package main
 
 import (
-	"log"
+	"fmt"
+	"os"
 
 	"github.com/operator-framework/combo/cmd"
 )
 
 func main() {
 	if err := cmd.Execute(); err != nil {
-		log.Fatal(err)
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
 	}
 }
--- a/pkg/combination/combination.go
+++ b/pkg/combination/combination.go
@@ -69,10 +69,6 @@ func (cs *streamImp) Next(ctx context.Context) (map[string]string, error) {
 		}
 	}
 
-	if cs.solveAhead && !cs.solved {
-		cs.solve()
-	}
-
 	if len(cs.combinations) == 0 {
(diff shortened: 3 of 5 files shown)


pkg/combination/combination.go
```diff
@@ -69,10 +69,6 @@ func (cs *streamImp) Next(ctx context.Context) (map[string]string, error) {
 		}
 	}
 
-	if cs.solveAhead && !cs.solved {
-		cs.solve()
-	}
```
timflannagan, line 74: Any reason this was removed? The linter complained about it?

main.go
```diff
@@ -1,9 +1,13 @@
...
 import (
+	"log"
+
 	"github.com/operator-framework/combo/cmd"
 )
 
 func main() {
-	cmd.Execute()
+	if err := cmd.Execute(); err != nil {
+		log.Fatal(err)
```
timflannagan, line 11: This is purely a nit so feel free to ignore: I'm personally a fan of this error handling at the main function level:

```go
if err := cmd.Execute(); err != nil {
    fmt.Fprintln(os.Stderr, err)
    os.Exit(1)
}
```

You likely get the same functionality under-the-hood with log.Fatal so that suggestion likely falls under the personal preference realm.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,21 @@
+name: ci
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    paths-ignore:
+    - '**/*.md'
+jobs:
+  unit:
```
timflannagan, line 10: Copy/paste error: you'll need a unique job identifier.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v2
+      - name: Install Go 1.17
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.17'
+      - name: Run unit tests
```
timflannagan, line 19: Copy/paste error: you'll want something like "Run sanity checks", or something along those lines.

pkg/combination/combination.go
tylerslaton: It was some left over code I forgot to remove in a previous PR it looks like and the linter caught the error wasn't checked. If you look at the line above it is doing the same thing but checking the error.

.github/workflows/lint.yaml
tylerslaton: Yep, good catch. Fixing.

main.go
tylerslaton: Hm, interesting. Any reason one approach is preferable to the other?

timflannagan (approved):

PR #23, opened 2021-11-05
cmd/eval: Update the eval command help description

timflannagan:
Update the eval sub-command for the cobra binary:

- Deincrement the example used in the long description
- Ensure the cmd.Use exposes what type of argument is required
- Make a note that the --replacements flag is a required flag in the
  long description

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -15,14 +15,18 @@ import (
 var (
 	replacements map[string]string
 	evalCmd      = &cobra.Command{
-		Use:   "eval",
+		Use:   "eval [file]",
 		Short: "Evaluate the combinations for a file at the given path",
-		Long: `Evaluate the combinations for a file at the given path. The file provided must be valid YAML.
+		Long: `
+Evaluate the combinations for a file at the given path. The file provided must be valid YAML.
 
-	Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
+Note: the combo binary requires the --replacement flag to be explicitly set.
+
+The replacements flag allows users to specify a series of key value pairs in the form of KEY=VALUE.
+
+Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
 	`,
 		RunE: run,
-		Args: cobra.ExactArgs(1),
 	}
 )
 


cmd/eval.go
```diff
@@ -15,14 +15,18 @@ import (
...
-		Long: `Evaluate the combinations for a file at the given path. The file provided must be valid YAML.
+		Long: `
+Evaluate the combinations for a file at the given path. The file provided must be valid YAML.
 
-	Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
+Note: the combo binary requires the --replacement flag to be explicitly set.
+
+The replacements flag allows users to specify a series of key value pairs in the form of KEY=VALUE.
+
+Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
```
timflannagan, line 27: I de-indented (not entirely that's a word) this example as it was rendering strangely for me locally:

```bash
$ ./combo eval --help
Evaluate the combinations for a file at the given path. The file provided must be valid YAML.

	Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
...
```

cmd/eval.go
```diff
@@ -15,14 +15,18 @@ import (
 var (
 	replacements map[string]string
 	evalCmd      = &cobra.Command{
-		Use:   "eval",
+		Use:   "eval [file]",
```
timflannagan, line 18: Open to suggestions on how to best expose what argument we are expecting. Looking at the current version of this sub-command on the main branch locally:

```bash
$ ./combo eval --help
...
Usage:
  combo eval [flags]
```

It's not entirely clear that we expect/require an argument to be passed. Adding the "eval [file]" results in the new executable help format:

```bash
$ ./combo eval --help
...
Usage:
  combo eval [file] [flags]
```

Which at least makes an attempt that an argument is required, on top of the example that's provided in the long description of the sub-command. Any thoughts?

cmd/eval.go
tylerslaton: Yeah that's odd, good catch.

cmd/eval.go
tylerslaton: I like adding the `[file]` here personally. It makes it pretty clear what needs to be passed when combined with the long description like you said.

tylerslaton (approved):
Good catch!

PR #24, opened 2021-11-09
initial combo controller

### LINKED ISSUE #2: basic combo controller
As a user, I want a way to create and evaluate templates on a Kubernetes cluster.

Acceptance Criteria:

- `Template` k8s API to define a template and its parameters
- `Combinations` k8s API to evaluate a template and inspect the result
- [level triggered](https://hackernoon.com/level-triggering-and-reconciliation-in-kubernetes-1f17fe30333d) evaluation; i.e. changes to template/combination specs cause status to be recalculated
- k8s APIs defined by CRDs
- controllers implemented via [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime)
- behavior is documented

tylerslaton:
# Summary
Adding the manifests, controller boilerplate, and updated Makefile target for installing the Combo controller. The actual reconciliation logic will be addressed in #33 

# Additions
- Overhauled `make install` to install CRDs before applying the new Deployment explicitly
- Controller package with boilerplate (reconciliation coming soon)
- Manifests directory:
  - Generated CRDs
  - Controller deployment YAML

Closes #2


tylerslaton:
Looks like the `sanity` action timed out while trying to build. Is there any way that we can speed up compilation? My guess is that it downloading all of the dependencies is taking all the time.

cmd/run.go
```diff
@@ -0,0 +1,44 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
+)
+
+func init() {
+	rootCmd.AddCommand(runCmd)
+}
+
+// controllerCmd represents the controller command
```
awgreene, line 15: Description is off.

cmd/run.go
```diff
@@ -0,0 +1,44 @@
...
+var runCmd = &cobra.Command{
+	Use:   "run",
+	Short: "Run Combo as a controller on the cluster",
+	Long:  `add long description`,
```
awgreene, line 19: Add a description.

main.go
```diff
@@ -5,10 +5,12 @@ import (
 	"os"
 
 	"github.com/operator-framework/combo/cmd"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
 func main() {
-	if err := cmd.Execute(); err != nil {
+	log := zap.New()
+	if err := cmd.Execute(log); err != nil {
```
awgreene, line 13:
```suggestion
	if err := cmd.Execute(zap.New()); err != nil {
```

manifests/deploy/cluster_role.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:controller:combo
+rules:
+- apiGroups: ["combo.io"]
+  resources: ["*"]
+  verbs: ["*"]
```
awgreene, line 8: Nit: Newline.

awgreene: Scope the permissions of the controller. For example, I don't believe that this controller needs to delete either of the combo.io CRs.

manifests/deploy/cluster_role_binding.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: olm-operator-binding-olm
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:controller:combo
+subjects:
+- kind: ServiceAccount
+  name: combo-account
+  namespace: combo
```
awgreene, line 12: Nit: Newline.

manifests/deploy/cluster_role_binding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: olm-operator-binding-olm
```
awgreene, line 4: Seems like a copy/paste issue.

manifests/deploy/cluster_role.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:controller:combo
```
awgreene, line 4: Typically I've seen roles, roleBindings, serviceAccounts, and the deployment share the same name, could we call each of these resources `combo-operator`?

manifests/deploy/deployment.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: combo
+  name: combo-deployment
```
awgreene, line 5: `combo-operator`.

manifests/deploy/deployment.yaml
```diff
@@ -0,0 +1,24 @@
...
+  labels:
+    app: combo
```
awgreene, line 7: `combo-operator`

manifests/deploy/deployment.yaml
```diff
@@ -0,0 +1,24 @@
...
+      labels:
+        app: combo
+    spec:
+      serviceAccountName: combo-account
+      containers:
+      - name: combo
+        image: quay.io/operator-framework/combo:latest
+        imagePullPolicy: Never
+        ports:
+        - containerPort: 8080
```
awgreene, line 24: Nit: Newline.

manifests/deploy/deployment.yaml
```diff
@@ -0,0 +1,24 @@
...
+    app: combo
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: combo
+  template:
+    metadata:
+      labels:
+        app: combo
```
awgreene, line 16: `combo-operator`

manifests/deploy/service_account.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  namespace: combo
+  name: combo-account
```
awgreene, line 5: Nit: newline.

pkg/controller/combo_controller.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+type combinationController struct {
+	client.Client
+
+	log logr.Logger
+}
+
+func (c *combinationController) manageWith(mgr ctrl.Manager) error {
+	// Create multiple controllers for resource types that require automatic adoption
```
awgreene, line 22: There is only one controller.

pkg/controller/combo_controller.go
```diff
@@ -0,0 +1,45 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/operator-framework/combo/api/v1alpha1"
+)
+
+type combinationController struct {
+	client.Client
+
+	log logr.Logger
```
awgreene, line 18:
```suggestion
	client.Client
	log logr.Logger
```

pkg/controller/combo_controller.go
```diff
@@ -0,0 +1,45 @@
...
+	// Create multiple controllers for resource types that require automatic adoption
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.Combination{}).
+		Complete(c)
+}
+
+func (c *combinationController) Reconcile(ctx context.Context, req ctrl.Request) (reconcile.Result, error) {
+	// Set up a convenient log object so we don't have to type request over and over again
+	log := c.log.WithValues("request", req)
+	log.Info("reconciling bundle")
```
awgreene, line 31:
```suggestion
	log.Info("reconciling combination")
```

pkg/controller/template_controller.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/operator-framework/combo/api/v1alpha1"
+)
+
+// +kubebuilder:rbac:groups=core.rukpak.io,resources=provisionerclasses,verbs=get;list;watch;create;update;patch;delete
+
+type templateController struct {
+	client.Client
+
+	log logr.Logger
+}
```
awgreene, line 20:
```suggestion
// +kubebuilder:rbac:groups=core.rukpak.io,resources=provisionerclasses,verbs=get;list;watch;create;update;patch;delete
type templateController struct {
	client.Client
	log logr.Logger
}
```

awgreene (changes requested):
Nice work @tylerslaton, had some feedback but this review primarily consisted of nits.

Makefile
```diff
@@ -32,8 +32,16 @@ format: ## Format the source code
...
 build-cli:
-	$(Q)go build -o combo
+	$(Q)go build -o bin/combo
+
+# TODO: Version container
+build-container: build-cli
+	docker build . --tag=quay.io/operator-framework/combo
+
+load-container: build-container
+	kind load docker-image quay.io/operator-framework/combo
```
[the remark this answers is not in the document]
awgreene: Same applies to the other `install, uninstall, fresh-install,etc..` targets. I recognize that this is super nitty and I don't feel the need to remove the targets if you like including them.

Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM golang:1.17
+WORKDIR /
+COPY . .
```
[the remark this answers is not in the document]
tylerslaton: Yep, good call. I'll tidy this up.

Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM golang:1.17
```
timflannagan, line 1: We may want an issue solely for bumping to Go 1.17 throughout the repository.

Dockerfile
```diff
@@ -0,0 +1,5 @@
...
+WORKDIR /
+COPY . .
+RUN make build-cli
+CMD ./bin/combo run
```
timflannagan, line 5: I think we'd want to have a multi-stage Dockerfile here: build the CLI in one stage, and in the second copy that binary over so the overall image size is lower. This can likely be done in a follow-up though.

Makefile
```diff
@@ -32,8 +32,16 @@ format: ## Format the source code
...
 	$(Q)go mod tidy
 
+# TODO: Disable CGO? Define specific tags? Do a from-scratch and copy pre-built binary. Static vs Dynamic linking
 build-cli:
-	$(Q)go build -o combo
+	$(Q)go build -o bin/combo
+
+# TODO: Version container
+build-container: build-cli
+	docker build . --tag=quay.io/operator-framework/combo
```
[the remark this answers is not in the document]
timflannagan: Yeah you can define a Makefile variable that defaults to quay.io/operator-framework/combo that a user is responsible for explicitly overriding when running this target at the command line.

Makefile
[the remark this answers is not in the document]
timflannagan: I'm personally a fan of housing install/uninstall type targets, but the load-container target seems like it belongs in a dev Makefile more than anything - including that target here isn't blocking imo.

Dockerfile
tylerslaton: Is this already done? That's the version defined in the `go.mod` currently.

Makefile
```diff
@@ -52,7 +60,16 @@ verify: tidy generate format
...
+	kubectl apply -f manifests/crds && kubectl apply -f manifests/deploy
+
+install-fresh: generate load-container install
+
+uninstall:
+	kubectl delete -f manifests/deploy && kubectl delete -f manifests/crds
+
+reinstall: uninstall install
+
+reinstall-fresh: uninstall install-fresh
```
timflannagan, line 72: I personally don't see a ton of value including these types of targets but I'm open to opinions that have a preference towards a reinstall-type target.

timflannagan (commented):
Just did a quick breeze through - thanks for tackling this initial scaffolding of the controller. I think it might be easier to push through this PR separating some of the api/manifest changes into the their own PR. I think the Makefile + introduction of a root Dockerfile can be separated into their own PR as well. Any thoughts?

Dockerfile
timflannagan: Yeah whoops you're right (I should've doubled check my bad :laughing:)

Makefile
tylerslaton: I mainly like reinstall types of targets since you just run one command to get everything related to the application up-to-date. That being said, I'm not strongly opinionated here, it was something I was using a lot while developing to make sure everything worked together but I could be convinced to not have it included in the repo.

Makefile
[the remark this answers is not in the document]
tylerslaton: Good call. Batch of changes for this and more coming through now.

tylerslaton:
> Just did a quick breeze through - thanks for tackling this initial scaffolding of the controller. I think it might be easier to push through this PR separating some of the api/manifest changes into the their own PR. I think the Makefile + introduction of a root Dockerfile can be separated into their own PR as well. Any thoughts?

I like this idea. I'll go ahead and break those changes out to make this more digestible.

Dockerfile
tylerslaton: Going to go ahead and break that into a separate PR.

Dockerfile
```diff
@@ -24,4 +24,4 @@ COPY --from=builder /build/bin/combo .
 
 EXPOSE 8080
 
-CMD ["/bin/combo"]
\ No newline at end of file
+CMD ["/bin/combo", "run"]
```
timflannagan, line 27: I think we should hold off changing the Dockerfile here and instead, handle this at the Deployment level by specify the command(s) explicitly. This should help alleviate any failures during the build workflow when attempting to run the newly built container image as the newly build container won't be expecting a cluster to be running (like the run command will need).

cmd/eval.go
```diff
@@ -88,3 +36,51 @@ func validateFile(file []byte) error {
 	var holder interface{}
 	return yaml.Unmarshal(file, &holder)
 }
+
+var (
+	replacements map[string]string
```
timflannagan, line 41: nit: Is a global variable needed here? Can this be defined in the RunE inline?

cmd/root_test.go
```diff
@@ -4,7 +4,7 @@ import "testing"
 
 func TestRoot(t *testing.T) {
 	t.Run("Execute", func(t *testing.T) {
-		if err := Execute(); err != nil {
+		if err := Execute(nil); err != nil {
```
timflannagan, line 7: nit: We should use a mock logger instead of nil here

manifests/deploy/deployment.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: combo
```
timflannagan, line 4: If we're going to be hardcoding the namespace with these manifests, we should introduce an explicit Namespace YAML manifest as well.

manifests/deploy/deployment.yaml
```diff
@@ -0,0 +1,24 @@
...
+  template:
+    metadata:
+      labels:
+        app: combo-operator
+    spec:
+      serviceAccountName: combo-operator
+      containers:
+      - name: combo
+        image: quay.io/operator-framework/combo:latest
+        imagePullPolicy: Never
```
timflannagan, line 22: By default, the imagePullPolicy is set to 'Always' implicitly anytime the `latest` tag is present. In any case, we should avoid using the 'Never' policy here and stick with the implicit 'Always' or specify 'IfNotPresent'.

manifests/deploy/deployment.yaml
```diff
@@ -0,0 +1,24 @@
...
+      app: combo-operator
+  template:
+    metadata:
+      labels:
+        app: combo-operator
+    spec:
+      serviceAccountName: combo-operator
+      containers:
+      - name: combo
+        image: quay.io/operator-framework/combo:latest
```
timflannagan, line 21: We can specify the `run` sub-command here:

```yaml
command:
- /bin/combo
- run
...
````

**Review by timflannagan — COMMENTED** [review]
Quick review - mainly on the list of YAML manifests being introduced - and overall, this looks like it's in a reasonable state. I held off reviewing any of the controller logic introduced here in case you're still iterating on that.

File: .github/workflows/build.yaml [hunk 29]
```diff
@@ -25,6 +25,4 @@ jobs:
       - name: Build the container
         run: |
           make build-container
-      - name: Run the container
-        run: |
-          docker run quay.io/operator-framework/combo:dev
+
```
timflannagan, line 28: I had mentioned this in https://github.com/operator-framework/combo/pull/24/files#r747829764 but it looks like I was reviewing a potential out-of-date diff at that time - I'd like to keep this check to avoid regressing on the overall container image being unable to run successfully.

Dockerfile
tylerslaton: Updated!

manifests/deploy/deployment.yaml
tylerslaton: Set a `IfNotPresent` here so that you can install locally and use a kind-loaded image without needing to pull.

.github/workflows/build.yaml
tylerslaton: Brought this back now!

cmd/eval.go
tylerslaton: Did some logic to remove this global variable. Makes things feel cleaner, good call.

manifests/deploy/deployment.yaml
tylerslaton: Went ahead and added a `namespace.yaml`. Also updated the Makefile targets to create it before the controller.

cmd/root_test.go
tylerslaton: Mind taking a look at the mock I made?

Makefile
```diff
@@ -56,5 +56,15 @@ test: test-unit ## Run the tests
 test-unit: ## Run the unit tests
 	$(Q)go test -count=1 -short ./...
 
+deploy: generate ## Deploy the Combo operator to the current cluster
+	kubectl apply -f manifests/crds			  && \
+	kubectl apply -f manifests/namespace.yaml && \
+	kubectl apply -f manifests/deploy
```
awgreene, line 62: There is no need to call `kubectl apply` multiple times. If you want to apply all files in the `/manifests` dir you could use:
```
$  k apply --recursive -f manifests
customresourcedefinition.apiextensions.k8s.io/combinations.combo.io created
customresourcedefinition.apiextensions.k8s.io/templates.combo.io created
clusterrole.rbac.authorization.k8s.io/combo-operator created
clusterrolebinding.rbac.authorization.k8s.io/combo-operator created
deployment.apps/combo-operator created
serviceaccount/combo-operator created
namespace/combo unchanged
```

Alternatively, if you want to apply the content at the existing paths, you can use this command:
```
k create -f manifests/namespace.yaml -f manifests/crds -f manifests/deploy
namespace/combo created
customresourcedefinition.apiextensions.k8s.io/combinations.combo.io created
customresourcedefinition.apiextensions.k8s.io/templates.combo.io created
clusterrole.rbac.authorization.k8s.io/combo-operator created
clusterrolebinding.rbac.authorization.k8s.io/combo-operator created
deployment.apps/combo-operator created
serviceaccount/combo-operator created
```

awgreene: This advice applies here as well: https://github.com/operator-framework/combo/pull/24/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R64-R67

Makefile
```diff
@@ -56,5 +56,15 @@ test: test-unit ## Run the tests
...
+
+teardown: ## Teardown the Combo operator to the current cluster
+	kubectl delete -f manifests/deploy		   && \
+	kubectl delete -f manifests/namespace.yaml && \
+	kubectl delete -f manifests/crds
+
 # Binary builds
-GO_BUILD := $(Q)go build
+GO_BUILD := $(Q)go build
```
awgreene, line 70: Nit: Newline. This advice applies in all instances, the reasoning behind this advice can be found [here](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline).

cmd/eval.go
```diff
@@ -88,3 +36,52 @@ func validateFile(file []byte) error {
...
+			generatedFile, err := generator.Evaluate(ctx, string(file), combinations)
+			if err != nil {
+				return err
+			}
+
+			if err := validateFile([]byte(generatedFile)); err != nil {
+				return fmt.Errorf("failed to validate file generated: %w", err)
+			}
+
+			fmt.Println(generatedFile)
```
awgreene, line 82: I am going to assume that the actual logic of the controller in which the generated resources are applied on cluster will appear in a follow up PR. As such, I will not be reviewing the controller logic at this time.

pkg/controller/combo_controller.go
awgreene: This was not addressed.

pkg/controller/controller.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+	}
+
+	return nil
+}
+
+type Controller struct {
+	client.Client
+
+	log     logr.Logger
```
awgreene, line 48:
```suggestion
	client.Client
	log     logr.Logger
```

pkg/controller/template_controller.go
```diff
@@ -0,0 +1,32 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/operator-framework/combo/api/v1alpha1"
+)
+
+// +kubebuilder:rbac:groups=core.rukpak.io,resources=provisionerclasses,verbs=get;list;watch;create;update;patch;delete
+
+type templateController struct {
```
awgreene, line 16:
```suggestion
// +kubebuilder:rbac:groups=core.rukpak.io,resources=provisionerclasses,verbs=get;list;watch;create;update;patch;delete
type templateController struct {
```

awgreene (changes requested):
Nice work @tylerslaton, had a few suggestions.

Makefile
tylerslaton: The main reason I avoided doing this is because the order of these being created matters:

```
customresourcedefinition.apiextensions.k8s.io/combinations.combo.io created
customresourcedefinition.apiextensions.k8s.io/templates.combo.io created
clusterrole.rbac.authorization.k8s.io/combo-operator created
clusterrolebinding.rbac.authorization.k8s.io/combo-operator created
namespace/combo created
Error from server (NotFound): error when creating "manifests/deploy/deployment.yaml": namespaces "combo" not found
Error from server (NotFound): error when creating "manifests/deploy/service_account.yaml": namespaces "combo" not found
```
Is there a way to get around this problem with this recursive solution?

cmd/root_test.go
```diff
@@ -1,10 +1,23 @@
...
+func (ml mockLogger) Info(msg string, keysAndValues ...interface{})             {}
+func (ml mockLogger) Enabled() bool                                             { return false }
+func (ml mockLogger) WithName(name string) logr.Logger                          { return mockLogger{} }
+func (ml mockLogger) WithValues(keysAndValues ...interface{}) logr.Logger       { return mockLogger{} }
+func (ml mockLogger) V(level int) logr.Logger                                   { return mockLogger{} }
 
 func TestRoot(t *testing.T) {
 	t.Run("Execute", func(t *testing.T) {
-		if err := Execute(); err != nil {
+		if err := Execute(mockLogger{}); err != nil {
```
timflannagan, line 20: I think my phrasing around "mock" logger likely wasn't clear so that's my bad - what I had in mind was passing the [`NewTestingLogger(t)`](https://pkg.go.dev/github.com/go-logr/logr/testing#NewTestLogger) function.

pkg/controller/template_controller.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/go-logr/logr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/operator-framework/combo/api/v1alpha1"
+)
+
+// +kubebuilder:rbac:groups=core.rukpak.io,resources=provisionerclasses,verbs=get;list;watch;create;update;patch;delete
```
timflannagan, line 14: This is a copy-paste left over from the rukpak repository. This isn't super important currently, but we'll want to update this to use the combo group and the template custom resource (vs. the provisionerclass resource currently specified).

pkg/controller/controller.go
```diff
@@ -0,0 +1,77 @@
...
+		log:    log,
+		managed: manageables{
+			&templateController{
+				Client: cli,
+				log:    log.WithValues("controller", "template"),
+			},
+			&combinationController{
+				Client: cli,
+				log:    log.WithValues("controller", "combination"),
+			},
```
timflannagan, line 64: I find this kind of structure a bit awkward to develop on personally, especially when the controller is expecting extra configuration (e.g. uses the options pattern) that cannot be handled by defaulting values, but this is fine for now.

pkg/controller/combo_controller.go
tylerslaton: Oops, I guess it got lost in some rebasing I was doing a bit back. Fixed now!

Makefile
timflannagan: You can get around that error by ensuring the namespace YAML manifest is lexicographically lesser than the manifests that require it (basically any of the namespaced-scoped controller resources).

timflannagan: More concretely, using prefixes:

- 01_namespace.yaml
- 02_deployment.yaml
- 02_serviceaccount.yaml
...

timflannagan: Another alternative might be to remove the namespace parameter entirely, and let kubelet apply the resources in whatever the current (or default) context has specified for a namespace. The only caveat would be RBAC that wants an explicit namespace parameter for namespace-scoped tenants, but I don't know off-hand whether that's a required field.

tylerslaton: The namespace is required for the RBAC so I believe we'd have to go with the lexicographical ordering here.

awgreene: I would suggest one of the following solutions:
1. Change the `/manifests/namespace.yaml` to `/manifests/00_namespace.yaml` so it is applied before either of the dirs that you created.
2. Put all the yaml in `/manifests` and add prefixes to control ordering of application.

Personally, I lean towards option 2.

timflannagan: I think I largely lean towards option 2 as well but I'd like to propose another alternative: we could just settle on a Makefile target that simply runs a bash wrapper script, utilizing envsubst to provide an absolutely basic templating layer without introducing additional tooling (helm, kustomize, etc.) while we iterate further. That way, we can at least have control over the $NAMESPACE that gets created by overriding the default value locally. Any thoughts?

awgreene: Seems reasonable, but can that be reported as an issue and addressed at a later date?

awgreene: Sorry - I had only glimpsed at your comment. You are effectively proposing that the namespace should be configurable via a bash script, and that the bash script should be in charge of ordering the order that resources are applied. That seems useful to get in earlier rather than later.

awgreene: Hmm - could we use combo to substitute the $NAMESPACE variable in the manifests being generated?

timflannagan: Let's just settle on introduced a deflated set of manifests that use numeric prefixes to convey ordering. We can iterate going forward as we expand the controller logic anyways. In case anyone is interested, this is a real simple, concrete example of what I had suggested: https://github.com/timflannagan/k8s-scripts/blob/master/scripts/create-nfs-provisioner.sh#L25-L27.

cmd/root_test.go
tylerslaton: Fixed this up, my bad. Should have asked for clarification.

pkg/controller/template_controller.go
tylerslaton: This may actually not be needed at all as these permissions are all provided via the `Role` and `Rolebinding` for in the `deployment.yaml`. What do you think though?

timflannagan: Those are used to help inform resource generation: https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/kubebuilder_annotation.md. If we're supplying our own RBAC definitions vs. using what kubebuilder scaffolded out for us, I think we can drop the RBAC annotations entirely from the controller(s) packages.

tylerslaton: Got it, in that case I'll remove this comment.

Makefile
tylerslaton: Okay cool, I've got this setup. Let me know what you think in the next push.

pkg/controller/controller.go
tylerslaton: I could see this being refactored to utilize a factory function. Let me take a stab at that.

cmd/eval.go
```diff
@@ -88,3 +36,52 @@ func validateFile(file []byte) error {
...
+Note: the combo binary requires the --replacement flag to be explicitly set.
+
+The replacements flag allows users to specify a series of key value pairs in the form of KEY=VALUES.
+
+Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
+	`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			replacements, _ := cmd.Flags().GetStringToString("replacements")
+
+			file, err := os.ReadFile(args[0])
```
timflannagan, line 56: This will produce a panic if no input file has been specified. We can enforce this at the command-level through the `ExactArgs` parameter.

cmd/eval.go
```diff
@@ -88,3 +36,52 @@ func validateFile(file []byte) error {
...
+Evaluate the combinations for a file at the given path. The file provided must be valid YAML.
+
+Note: the combo binary requires the --replacement flag to be explicitly set.
+
+The replacements flag allows users to specify a series of key value pairs in the form of KEY=VALUES.
+
+Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
+	`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			replacements, _ := cmd.Flags().GetStringToString("replacements")
```
timflannagan, line 54: Let's handle the error here vs. suppressing it.

pkg/controller/controller.go
tylerslaton: This could be an alternative that I came up with to get around the no logical default issue:

```go
managed: manageables{
    template.NewController(cli, log.WithValues("contoller", "template")),
    combo.NewController(cli, log.WithValues("contoller", "combination")),
},
```

pkg/controller/template_controller.go
```diff
@@ -0,0 +1,30 @@
...
+func (t *templateController) manageWith(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.Template{}).
+		Complete(t)
+}
+
+func (t *templateController) Reconcile(ctx context.Context, req ctrl.Request) (reconcile.Result, error) {
+	// Set up a convenient log object so we don't have to type request over and over again
+	log := t.log.WithValues("request", req)
+	log.V(1).Info("reconciling template")
```
timflannagan, line 28: Any idea on how we'd be able to configure t.log's logging level using the manageable's implementation?

cmd/eval.go
```diff
@@ -88,3 +36,52 @@ func validateFile(file []byte) error {
...
+
+			file, err := os.ReadFile(args[0])
+			if err != nil {
+				return err
+			}
+
+			if err := validateFile(file); err != nil {
```
timflannagan, line 61: Any gauge on the UX for an existing, empty file here? Produce a warning message, or an error message indicating that there's no substitution to be performed?

pkg/controller/controller.go
timflannagan: That looks reasonable but I think we just hold off on refactoring for now and if the implementation is too cumbersome when trying to configure multiple controllers, we can take a stab then.

cmd/eval.go
```diff
@@ -88,3 +36,52 @@ func validateFile(file []byte) error {
...
+
+			combinations := combination.NewStream(
+				combination.WithArgs(formatReplacements(replacements)),
+				combination.WithSolveAhead(),
+			)
+
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			generatedFile, err := generator.Evaluate(ctx, string(file), combinations)
```
timflannagan, line 73: This is out-of-scope for this PR, but the usage of `string(file)` here seems like it warrants some sort of refactoring to be agnostic to the input (maybe an io.Reader?). Diving into that Evaluate function, it looks like that logic for handling multi-document YAML manifests could be deflated as an options parameter to the NewStream constructor? Note: I hadn't dived too deep into the actual codebase, so I'm likely missing a good chunk of context, but something that I had noticed that we might be able to improve in the future.

cmd/eval.go
timflannagan: I'm fine with marking this out-of-scope for this PR if you want to open a follow-up issue/PR/etc. to handle that edge case.

tylerslaton: I like this, let me change up that valdiateFile function to do this.

tylerslaton:
```shell
go run main.go eval -r "TEST=1" test.yaml
Error: failed to validate file specified: empty file
Usage:
  combo eval [file] [flags]

Flags:
  -h, --help                          help for eval
  -r, --replacements stringToString   Key value pair of comma delimited values. Example: 'NAMESPACE=foo,bar' (default [])

failed to validate file specified: empty file
exit status 1
```

manifests/00_namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: combo
```
timflannagan, line 4: nit: EOF newline

cmd/eval.go
timflannagan: Cool, that works for me for now. We'll want to iron out that UX eventually in the context of CI pipelines running the eval sub-command.

cmd/eval.go
tylerslaton: You're right that the `string(file)` here is probably an indicator that the interface for that function should be updated. I could see a situation where we use the `io.Reader` instead to get around that wonkiness but yeah lets address that in the future.

pkg/controller/template_controller.go
tylerslaton: This is what I came up with:

```go
// controller.go
...
func (m manageables) manageWith(mgr ctrl.Manager) error {
	for _, man := range m {
		if m == nil {
			// Panic because this case indicates a bug
			panic("failed to register controller with manager: cannot be nil")
		}

		if err := man.manageWith(mgr, 1); err != nil {
			// Bail out if any controller fails to register with the manager
			return err
		}
	}

	return nil
}
...
```

```go
// combo_controller.go
...
func (c *combinationController) manageWith(mgr ctrl.Manager, version int) error {
	c.log = c.log.V(version)
	return ctrl.NewControllerManagedBy(mgr).
		For(&v1alpha1.Combination{}).
		Complete(c)
}
...
```
What do you think?

cmd/eval.go
```diff
@@ -86,5 +39,62 @@ func formatReplacements(replacementsInput map[string]string) map[string][]string
...
+			}
+
+			file, err := os.ReadFile(args[0])
+			if err != nil {
+				return fmt.Errorf("failed to read file specified: %w", err)
+			}
+
+			if err := validateFile(file); err != nil {
+				return fmt.Errorf("failed to validate file specified: %w", err)
+			}
```
timflannagan, line 76: (out-of-scope) I'd like to see us refactor this driver function at some point and ensure that validateFile is actually stating the file and that the file exists as minimal checks while we iterate further. That could mean we also get access to existing stdlib methods for checking for empty files:

```go
f, err := os.Stat(file)
// handle error + os.IsNotExists(err)
if f.Size() == 0 {
  return ErrEmptyFile
```
I haven't dived into whether there's a simpler way to do this, but we should be able to clean up the process of validating the file, and opening that file for reading as this project gets further into development.

cmd/eval.go
```diff
@@ -86,5 +39,62 @@ func formatReplacements(replacementsInput map[string]string) map[string][]string
 // validateFile is a simple wrapper to ensure the input/output of valid YAML
 func validateFile(file []byte) error {
 	var holder interface{}
-	return yaml.Unmarshal(file, &holder)
+	err := yaml.Unmarshal(file, &holder)
+	if holder == nil {
+		return ErrEmptyFile
+	}
+	return err
```
timflannagan, line 46: Thanks for updating - I think this is fine for now, but we should aim to clean this functionality up a bit while we iterate. See https://github.com/operator-framework/combo/pull/24/files#r749580994 for more information on one approach.

timflannagan: Also, we should add a test case for this, but we can do that as a follow-up.

timflannagan (approved):
I have a couple more comments, but I'm inclined to approve this PR as most of the functionality is already there, and the majority of the changes I'd like to see can be considered out-of-scope at this point, or changes that can be done as follow-ups as we continue iterating on this project. Thanks for driving much of the development on this project and splitting up this PR into individual ones to help reduce the review load for this initial controller scaffolding.

PR #25, opened 2021-11-10
Ignore the vendor directory

timflannagan:
Update the root .gitignore and ignore tracking any changes made to the
root vendor tree as we're not introducing that directory into source
control.


tylerslaton (approved):
👍🏻

PR #26, opened 2021-11-10
ci(sanity): vendor deps before running linter

tylerslaton:
# Summary
As more dependencies are getting added the linter is timing out when trying to download them before its run in the `sanity` action. As a result, this PR vendors the dependencies prior so that lint step is not responsible for downloading.

# Additions
- Vendor prior to running linter in `sanity` action
- Moved around targets to be more readable

### DIFF
--- a/.github/workflows/sanity.yaml
+++ b/.github/workflows/sanity.yaml
@@ -16,6 +16,7 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: '1.17'
+      - name: Vendor dependencies
+        run: go mod vendor
       - name: Run sanity checks
-        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/sanity.yaml
```diff
@@ -16,6 +16,7 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: '1.17'
+      - name: Vendor dependencies
+        run: go mod vendor
```
timflannagan, line 20: We may want to also do this throughout the other workflows at some point, but this is fine for now :+1:

timflannagan (approved):

PR #27, opened 2021-11-10
ci(sanity): add .golangci.yml with increased timeout for linter

tylerslaton:
Adding a golangci.yml and increasing the linter's timeout.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,3 @@
+run:
+  # Default timeout is 1m, up to give more room
+  timeout: 4m
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,3 +1,25 @@
 run:
   # Default timeout is 1m, up to give more room
   timeout: 4m
+
+linters:
+  enable:
+  - asciicheck
+  - deadcode
+  - depguard
+  - errorlint
+  - gofmt
+  - goimports
+  - importas
+  - prealloc
+  - nestif
+  - revive
+  - misspell
+  - stylecheck
+  - tparallel
+  - unconvert
+  - unparam
+  - whitespace
+
+output:
+  format: tab
--- a/pkg/generator/generator_test.go
+++ b/pkg/generator/generator_test.go
@@ -6,7 +6,7 @@ import (
 	"testing"
 
 	"github.com/operator-framework/combo/pkg/combination"
-	"github.com/operator-framework/combo/test/assets/generatorTestData"
+	testdata "github.com/operator-framework/combo/test/assets/generator"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -25,10 +25,10 @@ func TestEvaluate(t *testing.T) {
 	}{
 		{
 			name:     "can process a template",
-			template: generatorTestData.EvaluateInput,
+			template: testdata.EvaluateInput,
 			expected: expected{
 				err:        nil,
-				evaluation: generatorTestData.EvaluateOutput,
+				evaluation: testdata.EvaluateOutput,
 			},
 			combinations: combination.NewStream(
 				combination.WithArgs(map[string][]string{
@@ -66,8 +66,8 @@ func TestEvaluate(t *testing.T) {
 			assert.Equal(t, tt.expected.err, err)
 			require.ElementsMatch(
 				t,
-				strings.Split(string(tt.expected.evaluation), "---"),
-				strings.Split(string(evaluation), "---"),
+				strings.Split(tt.expected.evaluation, "---"),
+				strings.Split(evaluation, "---"),
 				"Document evaluations generated incorrectly")
 		})
 	}
--- a/test/assets/combination/testdata.go
+++ b/test/assets/combination/testdata.go
@@ -1,12 +1,12 @@
-package combinationTestData
+package testdata
 
-var CombinationInput map[string][]string = map[string][]string{
+var CombinationInput = map[string][]string{
 	"TEST1": {"foo", "bar"},
 	"TEST2": {"zip", "zap"},
 	"TEST3": {"bip", "bap"},
 }
 
-var CombinationOutput []map[string]string = []map[string]string{
+var CombinationOutput = []map[string]string{
 	{
 		"TEST1": "foo",
 		"TEST2": "zip",
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ generate: ## Generate code and manifests
 format: ## Format the source code
 	$(Q)go fmt ./...
 
-lint: ## Run golangci-lin
+lint: ## Run golangci-lint
 	$(Q)go run github.com/golangci/golangci-lint/cmd/golangci-lint run
 
 verify: tidy generate format lint ## Verify the current code generation and lint
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -37,7 +37,6 @@ func init() {
 		fmt.Fprintf(os.Stderr, "failed to initialize eval: %v", err)
 		os.Exit(1)
 	}
-
 }
 
 // run is used during the actual execution of the command to generate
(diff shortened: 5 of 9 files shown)


.golangci.yaml
```diff
@@ -0,0 +1,3 @@
+run:
+  # Default timeout is 1m, up to give more room
+  timeout: 4m
```
timflannagan, line 3: Can we also define a set of linters that we'd like to use? Are there any packages that need to be skipped (i.e. anything code generated)?

timflannagan: I was poking around [linter options](https://golangci-lint.run/usage/linters/) and came up with this initial list locally:

```yaml
run:
  # Default timeout is 1m, up to give more room
  timeout: 4m

linters:
  enable:
  - asciicheck
  - deadcode
  - depguard
  - errorlint
  - gofmt
  - goimports
  - importas
  - prealloc
  - nestif
  - revive
  - misspell
  - stylecheck
  - tparallel
  - unconvert
  - unparam
  - whitespace
```

WDYT?

timflannagan:
I ran that list of linters I threw in a comment locally and ran into some existing linting issues too:

```bash
$ golangci-lint run --out-format=tab
cmd/eval.go:40                                               whitespace  unnecessary trailing newline
cmd/eval.go:52:62                                            errorlint   non-wrapping format verb for fmt.Errorf. Use `%w` to format errors
cmd/eval.go:69:62                                            errorlint   non-wrapping format verb for fmt.Errorf. Use `%w` to format errors
pkg/combination/combination.go:113:2                         prealloc    Consider preallocating `arrays`
pkg/combination/combination.go:114:2                         prealloc    Consider preallocating `replacements`
pkg/combination/combination.go:44:41                         revive      unexported-return: exported func WithArgs returns unexported type combination.streamOption, which can be annoying to use
pkg/combination/combination.go:54:23                         revive      unexported-return: exported func WithSolveAhead returns unexported type combination.streamOption, which can be annoying to use
pkg/combination/combination_test.go:47:7                     errorlint   comparing with != will fail on wrapped errors. Use errors.Is to check for a specific error
pkg/combination/combination_test.go:70:8                     errorlint   comparing with != will fail on wrapped errors. Use errors.Is to check for a specific error
pkg/generator/generator.go:73:40                             misspell    `seperator` is a misspelling of `separator`
pkg/generator/generator_test.go:69:25                        unconvert   unnecessary conversion
pkg/generator/generator_test.go:70:25                        unconvert   unnecessary conversion
test/assets/combinationTestData/combinationTestData.go:3:22  revive      var-declaration: should omit type map[string][]string from declaration of var CombinationInput; it will be inferred from the right-hand side
test/assets/combinationTestData/combinationTestData.go:9:23  revive      var-declaration: should omit type []map[string]string from declaration of var CombinationOutput; it will be inferred from the right-hand side
test/assets/combinationTestData/combinationTestData.go:1:1   stylecheck  ST1003: should not use MixedCaps in package name; combinationTestData should be combinationtestdata
test/assets/generatorTestData/generatorTestData.go:3:19      revive      var-declaration: should omit type string from declaration of var EvaluateInput; it will be inferred from the right-hand side
test/assets/generatorTestData/generatorTestData.go:38:20     revive      var-declaration: should omit type string from declaration of var EvaluateOutput; it will be inferred from the right-hand side
test/assets/generatorTestData/generatorTestData.go:1:1       stylecheck  ST1003: should not use MixedCaps in package name; generatorTestData should be generatortestdata
```

.golangci.yaml
tylerslaton: So if I understand this correctly, these are additional linters that will run - right?

tylerslaton:
> I ran that list of linters I threw in a comment locally and ran into some existing linting issues too:
> 
> ```shell
> $ golangci-lint run --out-format=tab
> cmd/eval.go:40                                               whitespace  unnecessary trailing newline
> cmd/eval.go:52:62                                            errorlint   non-wrapping format verb for fmt.Errorf. Use `%w` to format errors
> cmd/eval.go:69:62                                            errorlint   non-wrapping format verb for fmt.Errorf. Use `%w` to format errors
> pkg/combination/combination.go:113:2                         prealloc    Consider preallocating `arrays`
> pkg/combination/combination.go:114:2                         prealloc    Consider preallocating `replacements`
> pkg/combination/combination.go:44:41                         revive      unexported-return: exported func WithArgs returns unexported type combination.streamOption, which can be annoying to use
> pkg/combination/combination.go:54:23                         revive      unexported-return: exported func WithSolveAhead returns unexported type combination.streamOption, which can be annoying to use
> pkg/combination/combination_test.go:47:7                     errorlint   comparing with != will fail on wrapped errors. Use errors.Is to check for a specific error
> pkg/combination/combination_test.go:70:8                     errorlint   comparing with != will fail on wrapped errors. Use errors.Is to check for a specific error
> pkg/generator/generator.go:73:40                             misspell    `seperator` is a misspelling of `separator`
> pkg/generator/generator_test.go:69:25                        unconvert   unnecessary conversion
> pkg/generator/generator_test.go:70:25                        unconvert   unnecessary conversion
> test/assets/combinationTestData/combinationTestData.go:3:22  revive      var-declaration: should omit type map[string][]string from declaration of var CombinationInput; it will be inferred from the right-hand side
> test/assets/combinationTestData/combinationTestData.go:9:23  revive      var-declaration: should omit type []map[string]string from declaration of var CombinationOutput; it will be inferred from the right-hand side
> test/assets/combinationTestData/combinationTestData.go:1:1   stylecheck  ST1003: should not use MixedCaps in package name; combinationTestData should be combinationtestdata
> test/assets/generatorTestData/generatorTestData.go:3:19      revive      var-declaration: should omit type string from declaration of var EvaluateInput; it will be inferred from the right-hand side
> test/assets/generatorTestData/generatorTestData.go:38:20     revive      var-declaration: should omit type string from declaration of var EvaluateOutput; it will be inferred from the right-hand side
> test/assets/generatorTestData/generatorTestData.go:1:1       stylecheck  ST1003: should not use MixedCaps in package name; generatorTestData should be generatortestdata
> ```

Would it make sense to include fixes to these into this PR as well?

timflannagan:
Yeah I'd include the fixes in a separate commit.

.golangci.yaml
tylerslaton: In terms of things we could skip, the only generated code I can think of is:

`api/v1alpha1/zz_generate.deepcopy.go`

timflannagan (approved):

pkg/generator/generator_test.go
```diff
@@ -6,7 +6,7 @@ import (
 	"testing"
 
 	"github.com/operator-framework/combo/pkg/combination"
-	"github.com/operator-framework/combo/test/assets/generatorTestData"
+	testdata "github.com/operator-framework/combo/test/assets/generator"
```
njhale, line 9: In go, the canonical directory and package name to use for test data is `testdata`.

test/assets/combination/testdata.go
```diff
@@ -1,12 +1,12 @@
-package combinationTestData
+package testdata
 
-var CombinationInput map[string][]string = map[string][]string{
+var CombinationInput = map[string][]string{
```
njhale, line 3: This test data is relatively small, do we really need to keep it in separate files? can we just define it up front, as literals, with the test cases themselves?

I'm usually all for DRY, but the one exception is test cases where:
- I don't want to _hide_ my expectations from the reader
- I don't want to create a shared dependency between tests/testcases; i.e. tweaking shared testdata usually always has unintended consequences, so we almost always end up writing functions to mutate that data -- invariably this leads to:
  - test cases being hard to grok; e.g. `input: withFoo(withBar(withBaz(mySharedInput), baz), bar), foo),`
  - more maintenance

njhale (commented):
Nice work!

I see this already merged, but I still thought up some comments that might be worth pursuing.

pkg/generator/generator_test.go
timflannagan: +1 here as well, despite this being out of scope for this PR (imo), I'd still like to see this get addressed sometime in the near future.

PR #29, opened 2021-11-10
.github/workflows: Notify pull request authors when workflows have failed

timflannagan:
Update the existing set of GHA workflows to notify PR authors when the workflow checks have failed. This job is triggered on non-draft PRs to avoid any rate limiting or spam notifications when iterating on WIP/draft PRs. Additionally, avoid using the same `ci` prefixed name. This can help better distinguish between individual workflows when interacting with the `actions` tab in the repositories github UI.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,4 +1,4 @@
-name: ci
+name: build
 on:
   push:
     branches:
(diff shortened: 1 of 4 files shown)


timflannagan:
Ah, it looks like this may not work as we primarily use forks for opening pull requests :(

It's also entirely possible I'm misunderstanding the error message and all of this would be resolved by merging this PR first.

In any case holding.

/hold

PR #30, opened 2021-11-10
feat: add Dockerfile

tylerslaton:
Adding a Dockerfile to build the Combo container. This is necessary to build the container defined in #24.


Makefile
```diff
@@ -38,9 +38,13 @@ lint: ## Run golangci-lin
 verify: tidy generate format lint ## Verify the current code generation and lint
 	git diff --exit-code
 
-build-cli:
+build-cli: ## Build the CLI binary
 	$(Q)go build -o combo
 
+TAG=quay.io/operator-framework/combo
```
timflannagan, line 44: nit: This is more of an "image" than tag. I'd recommend either allowing users to override this default image, or allowing them to override the default repository + tag. This is personal preference so feel free to ignore: can we move this variable definition to the top of the Makefile?

Makefile
```diff
@@ -38,9 +38,13 @@ lint: ## Run golangci-lin
...
+build-container: ## Build the Combo container
+	docker build . --tag=$(TAG)
```
timflannagan, line 46: nit: Explicitly specify the `-f Dockerfile` here?

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+COPY api api
+COPY cmd cmd
+COPY pkg pkg
+COPY test test
+COPY main.go main.go
+COPY Makefile Makefile
+COPY tools.go tools.go
+COPY go.mod go.mod
+COPY go.sum go.sum
```
timflannagan, line 13: Should we move this further up in the Dockerfile so we can utilize already cached layers and don't have to re-download the module cache anytime we make a code change (that doesn't affect dependencies):

```dockerfile
COPY go.mod go.mod
COPY go.sum go.sum
RUN go mod download

COPY api api
...
```

Any thoughts?

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+RUN make build-cli
+
+CMD ./bin/combo run
```
timflannagan, line 17: We'll need to update this to avoid specifying the `run` sub-command as it doesn't exist yet: how do you feel about the following structure instead:

```dockerfile
CMD ["/bin/combo"]
```

And the (future) controller deployment is responsible for providing the sub-command it wants to run as an argument?

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.17
+
+WORKDIR /
+
+COPY api api
+COPY cmd cmd
+COPY pkg pkg
+COPY test test
```
timflannagan, line 8: Are testing packages necessary to run the container? Can we remove their inclusion into the destination container image?

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+COPY main.go main.go
+COPY Makefile Makefile
+COPY tools.go tools.go
+COPY go.mod go.mod
+COPY go.sum go.sum
+
+RUN make build-cli
```
timflannagan, line 15: We'll want to migrate towards a multi-stage build process instead of a single stage to decrease the overall container image size. This should make it easier to reason about where this `make build-cli` throws the compiled executable as well.

timflannagan (commented):
Just a quick review of these changes - thanks for splitting this up from the original PR and tackling this overall work. I had a couple of questions/comments/etc. and the only blocking item from my side is the need for a multi-stage Dockerfile before merging.

It would be nice to have a GHA that also builds and uses this Dockerfile to ensure we're not regressing when iterating on this file in the future, or we're not introducing a Dockerfile that doesn't build and run correctly, but my current thinking is that I wouldn't block a PR without the inclusion of those tests.

Dockerfile
tylerslaton: Ahh, good idea. I like this.

Dockerfile
tylerslaton: Yeah, I wasn't entirely sure what to include here without the `run` subcommand existing but this is a better idea.

Dockerfile
tylerslaton: I went ahead and made a multi-stage process. What do you think?

Makefile
tylerslaton: Updated!

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+WORKDIR /build
+
+COPY go.mod go.mod
+COPY go.sum go.sum
+RUN go mod download
+COPY api api
+COPY cmd cmd
+COPY pkg pkg
+COPY main.go main.go
+COPY tools.go tools.go
```
timflannagan, line 12: nit: We shouldn't need tools.go in order to successfully run the container.

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+FROM golang:1.17-buster
+
+WORKDIR /
+
+COPY --from=builder /build/bin/combo /bin
+
+EXPOSE 8080
+
+CMD ["combo"]
```
timflannagan, line 25: I was playing around with this second stage locally:

```dockerfile
FROM gcr.io/distroless/static:nonroot
WORKDIR /bin
COPY --from=builder /build/bin/combo .
EXPOSE 8080

CMD ["/bin/combo"]
```

Any thoughts? In order to make it work with distroless final layers, you'd need to do something like the following:

```diff
-RUN make build-cli
-
-FROM golang:1.17-buster
-
-WORKDIR /
-
-COPY --from=builder /build/bin/combo /bin
+RUN CGO_ENABLED=0 go build -o ./bin/combo main.go
 
+FROM gcr.io/distroless/static:nonroot
+WORKDIR /bin
+COPY --from=builder /build/bin/combo .
 EXPOSE 8080
 
-CMD ["combo"]
\ No newline at end of file
+CMD ["/bin/combo"]
```

timflannagan (commented):
I think this is super close - had one suggestion about using the distroless image for the final stage, but other than that, the multi-stage layout looks reasonable to me.

Dockerfile
timflannagan: Another thing I just realized: do we need to "force" that combo binary to be rebuilt everytime so we're not caching it's output everytime we build the container? If so, we can either handle that as a `.FORCE: ...` target in the Makefile, or adding the `-a` flag to `go build -a -o ...`. Any thoughts?

.github/workflows/build.yaml
```diff
@@ -19,3 +19,6 @@ jobs:
       - name: Build the CLI
         run: |
           make build-cli
+      - name: Build the container
+        run: |
+          make build-container
```
timflannagan, line 24: Thanks - we'll want to run this container to verify it's actually working outside of the build process, but this is fine for now imo. We can also get that verification logic for free with the introduction of e2e tests, and I'm not super concerned with the container image breaking in the interim as it's not being used anywhere.

Dockerfile
tylerslaton: What would be the point of not caching it? I would think that there would be some logic in Docker/Podman to know when to and not to rebuild the binary. Maybe I'm making a wrong assumption there?

timflannagan: We might be fine then if we make the assumption that and change made to the codebase (e.g. api/cmd/pkg/etc directories) would result in a new binary anyways? This is a small detail though so I'm fine with resolving the "force" comment here.

.github/workflows/build.yaml
tylerslaton: This was fairly trivial to add so I just went ahead and did it! Also added the same "run" verification for the binary itself.

Dockerfile
tylerslaton: We may as well add it just to be explicit about it being new code.

timflannagan (approved):

PR #35, opened 2021-11-15
feat: add version subcommand

### LINKED ISSUE #22: add version subcommand to cli
# Story
As a user of the Combo CLI,
I want to be check what version I'm on,
So that I can debug if needed.

# Summary
Currently the Combo CLI has no version or version command associated with it. This is important for debugging/development purposes.

# Acceptance Criteria
- Version defined in the repo
- Subcommand to run `combo version`
- Unit test to validate functionality

tylerslaton:
# Summary
Adding a new subcommand for version. This grabs the current version in the new `COMBO_VERSION` file and loads it into the `version` package along with the SHA of the current Git commit. 

Example:

```shell
$ combo version
Combo version: v0.0.1
git commit: fefd778d2eb9e54b3ad58be0b26929e9fd369ee5
```

```shell
$ combo version --full
v0.0.1-fefd778d2eb9e54b3ad58be0b26929e9fd369ee5
```

# Additions
- New `version` subcommand with `--full` flag
- Makefile update to the `build` target to use `ldflags`

Closes #22


cmd/version.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"fmt"
+
+	"github.com/operator-framework/combo/pkg/version"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	rootCmd.AddCommand(versionCmd)
```
timflannagan, line 11: Should we update adding this sub-command to the root chain at the rootCmd call site instead? That way the definition of the rootCmd and it's child commands are all present centrally.

pkg/version/version.go
```diff
@@ -0,0 +1,19 @@
+package version
+
+import "fmt"
+
+// ComboVersion indicates what version of Combo the binary belongs to
+var ComboVersion string
+
+// GitCommit indicates which git commit the binary was built from
+var GitCommit string
```
timflannagan, line 9: WYDT about also tracking the Kubernetes version too?

pkg/version/version.go
```diff
@@ -0,0 +1,19 @@
+package version
+
+import "fmt"
+
+// ComboVersion indicates what version of Combo the binary belongs to
+var ComboVersion string
```
timflannagan, line 6: I think I'm more-or-less interested in the git tag + commit information for this version sub-command - wdyt? Maybe rename this variable to reflect the git tag here and update the root Makefile to set this ldflag explicitly if we're fine with that layout.

pkg/version/version.go
timflannagan, line 9: +1

cmd/version.go
```diff
@@ -0,0 +1,33 @@
...
+	versionCmd.Flags().BoolP("full", "f", false, "Returns the fully constructed version")
```
timflannagan, line 12: Any idea on why a full version string would be needed? For the most part, I would think that interacting with the version command's formatted string would be sufficient, but I'm curious whether you had something else in mind.

pkg/version/version.go
timflannagan: I realize there's no guarantee we're processing an unannotated tag, especially for HEAD commits, so maybe we need to use `git describe ...` here to grab lightweight tags.

cmd/version.go
tylerslaton: My main thought was that it could be used by some automation tools to confirm the current version in one line. That being said, I'm not too attached to keeping this.

COMBO_VERSION
```diff
@@ -0,0 +1 @@
+v0.0.1
```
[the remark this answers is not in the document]
tylerslaton: I also like the idea of moving away from a static file, would we use Git tags to keep track of this instead?

pkg/version/version.go
tylerslaton: Using Git tags here sounds like a much better solution than the static version file. I would like to see a little automation around bumping that version or at least making sure that we are bumping it manually with a tag each time. Have you done anything like that before? Probably out of the scope of this PR.

pkg/version/version.go
tylerslaton: Took a stab at tracking the Kubernetes version. What are your thoughts on the approach?

cmd/version.go
tylerslaton: So basically moving all of the `rootCmd.AddCommand()` to the `Execute()` function?

Something like this is what I'm seeing in my mind's eye from what you're saying, correct me if I'm wrong:

```go
// ./cmd/root.go
// Execute executes the root command.
func Execute(log logr.Logger) error {
	if log != nil {
		rootLog = log
	}
	
	// NEW =======
	rootCmd.AddCommand(versionCmd)
	rootCmd.AddCommand(evalCmd)
	rootCmd.AddCommand(runCmd)
        // ===========
        
	rootCmd.CompletionOptions.DisableDefaultCmd = true

	return rootCmd.Execute()
}

```

COMBO_VERSION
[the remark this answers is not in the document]
timflannagan: That would be my preference but I think a potential edge case is unnannotated tags - basically we're introducing new commits to a branch and those commits haven't been properly tagged. Maybe `git describe ...` has this functionality we're interested in?

Makefile
```diff
@@ -38,8 +43,10 @@ lint: ## Run golangci-lint
 verify: tidy generate format lint ## Verify the current code generation and lint
 	git diff --exit-code
 
-build-cli: ## Build the CLI binary
-	$(Q)go build -a -o ./bin/combo
+KUBERNETES_VERSION=v1alpha1
```
timflannagan, line 46: I think we're interested in the Kubernetes version we're importing as dependencies, so in this case, [0.22.2](https://github.com/operator-framework/combo/blob/main/go.mod#L12). We can likely just hardcode this value for now, and bump this version when bumping Kubernetes versions in the root go.mod, and just take the extra maintenance burden over time. wdyt?

cmd/version.go
timflannagan: That sounds reasonable to me assuming that's where the rootCmd is being defined, although that might just be a global variable currently (vs. we're using a constructor function to build up a rootCmd, returning a cobra.Command explicitly). We can iron out the command codebase structure over time, so let's settle on the Execute function for now.

cmd/version.go
timflannagan: I don't have a strong opinion either - I was mainly interested in reducing scope if we don't have an immediate use case for it.

pkg/version/version.go
timflannagan: Yeah, automation would be nice, but it's not something we have to iron out immediately. I'm like to get a structure in place before introducing automation when the manual process could be fairly trivial. Longer term, we'll likely want to introduce something like goreleaser to handle all of this for us anyways.

pkg/version/version.go
timflannagan: Replied in https://github.com/operator-framework/combo/pull/35/files#r750568999.

COMBO_VERSION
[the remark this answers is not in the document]
tylerslaton: Yeah, that would make it so we only get annotated tags like you mentioned. What would the process of tagging look like for now? Would we have to create the tag locally and then push it up everytime we merge?

cmd/version.go
tylerslaton: Makes sense. Updated.

Makefile
tylerslaton: What would you think about doing something like this so we can just automate grabbing that version?

```shell
go list -m all | grep k8s.io/apimachinery | awk '{print $2}'
```

We could just throw that into the `Makefile` and apply it as an `ldflag`.

timflannagan: I think either hardcoding it or adding that shell command you pasted is fine, although it's a bit hacky. Maybe add a TODO comment to iron that out if we're interested in that approach?

tylerslaton: Yeah, it's a bit hacky, I agree. Let's just hardcode it for now and come back to it if it gets to be too annoying.

pkg/version/version.go
tylerslaton: Okay, I've replaced the static file with a `git describe` in the Makefile. It also has a default version since we currently do not have any tags.

njhale (approved):

PR #36, opened 2021-11-17
.github/workflows: Enable workflow_dispatch events

timflannagan:
Update the existing set of CI-related workflows and enable the
workflow_dispatch functionality in order to have a first-class way of
triggering a workflow manually without having to entertain previous
workarounds, like opening a PR with an empty commit.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -6,6 +6,7 @@ on:
   pull_request:
     paths-ignore:
     - '**/*.md'
+  workflow_dispatch:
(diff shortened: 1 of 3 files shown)


timflannagan:
I somehow missed this when this functionality was added - check https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/ for more information.

tylerslaton (approved):
👏🏻 Nice, love to see it. Glad that adding something like this is super easy.

PR #37, opened 2021-11-17
.github/workflows: Avoid using the 'ci' workflow name for all CI-related actions

timflannagan:
Update any relevant CI-related workflows and avoid using the shared 'ci'
prefix for the workflow name in order better distinguish between
workflow(s) when interacting with the 'Actions' tab in the GitHub UI.

---
Screenshot of the previous behavior (using OLM as an example):

Screenshot from 2021-11-12 12-37-59
---

Screenshot of the new proposed behavior:

Screenshot from 2021-11-17 12-42-48

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,4 +1,4 @@
-name: ci
+name: build
 on:
   push:
     branches:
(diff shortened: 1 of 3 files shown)


tylerslaton (approved):
Nice UX change!

Issue #33, opened 2021-11-11
add reconciliation logic for controller


tylerslaton:
# Story
As a user of Combo,
I want to be able to have new Templates and Combinations reconciled,
So that I can get the Combinations of arguments created.

# Summary
As the controller currently stands (implemented in #24) it has sample reconciliation logic. This boilerplate should be replaced with actual code to handle the reconciliation as expected. 

# Acceptance Criteria

- [x] Combination Reconciliation
- [x] Combination e2e test
- [x] Template reconciliation
- [x] Template e2e test

PR #38, opened 2021-11-18
refactor(generator): update to be more compatible with being run in a controller

tylerslaton:
# Summary
The generator was due for a refactor. As I was finishing up the initial reconciliation logic for the combination controller it was evident that a different return format would be necessary. Along with this I updated parts of the package to be cleaner or more optimized. This PR is focused solely on refactoring the generator package to help with the cognitive load.

# Additions
- `template.go` file with split logic for handling and processing files with combinations
    - A test to go along with this file to validate logic
    - Optimized logic for factory function and `.with()` function
- A new `Generator` type for generating files
    - Evaluate method now returns a `[]string` to allow for more versatility when interacting with it.

# Questions
- Should the new interface designs be changed?
- Does it make sense to have the `Generator` type to utilize a `io.Reader` interface instead of a string?

Required for #33


pkg/generate/generate_test.go
```diff
@@ -13,19 +14,19 @@ import (
...
 		name         string
-		template     string
+		file         io.Reader
 		combinations combination.Stream
 		expected     expected
 	}{
 		{
-			name:     "can process a template",
-			template: testdata.EvaluateInput,
+			name: "can process a template",
```
perdasilva, line 28: should we also test error paths?

pkg/generate/template.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"errors"
+	"io"
+	"io/ioutil"
+	"regexp"
+	"strings"
+)
+
+var (
+	ErrCouldNotReadFile = errors.New("could not read file")
```
perdasilva, line 12: suggestion: factor errors out into their own package and re-use them here and in `eval`

pkg/generate/template.go
```diff
@@ -0,0 +1,67 @@
...
+			constructedTemplate.documents = append(constructedTemplate.documents, trimmedDoc)
+		}
+	}
+
+	return constructedTemplate, nil
+}
+
+// has determines if any of the documents for the template
+// contains the specified string.
+func (t template) has(find string) bool {
```
perdasilva, line 43: nit: `find` reads weird for me here. Would something like `str` or `searchString` or something like that?

pkg/generate/template.go
```diff
@@ -0,0 +1,67 @@
...
+	for _, document := range t.documents {
+		if document == find {
+			return true
+		}
+	}
+	return false
+}
+
+// with builds the template documents with the combination set specified
+func (t template) with(combo map[string]string, to template) template {
```
perdasilva, line 53: suggestion: since we're mutating `to` here would it make more sense to pass it in as a `*template` and remove the return value?

(not sure what the Go best practices would be here)

pkg/generate/template_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"io"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestHas(t *testing.T) {
```
perdasilva, line 11: maybe we should also exercise the error paths as well

pkg/generate/template_test.go
```diff
@@ -0,0 +1,114 @@
...
+				t.Fatal("error with test, not able to create template:", err)
+			}
+
+			actual := actualTemplate.has(tt.find)
+			require.Equal(t, tt.expected, actual)
+		})
+	}
+}
+
+func TestWith(t *testing.T) {
```
perdasilva, line 57: Do we care about cases where there's a key in the map but it doesn't exist in the template?

pkg/generate/generate.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"context"
+	"fmt"
+	"io"
+)
+
+type (
+	Generator interface {
+		Evaluate(ctx context.Context) ([]string, error)
```
perdasilva, line 11: A `Generator` with just an `Evaluate` method read weird to me. Would `Generate` fit better?

tylerslaton: Yeah, that reads a lot better.

pkg/generate/template.go
tylerslaton: Good idea, let me pencil out what that looks like.

pkg/generate/template.go
tylerslaton: This brings up a good point, I went ahead and refactored this to make a bit more sense given the mutation. This now works on the original template itself (optimizing memory a bit more) as well as being a bit easier to understand. The interface for interacting with this will now look like:

```go
file := strings.NewReader("hello: WORLD")

t, err := buildTemplate(file)
if err != nil {
    return err
}

t.with(combination)

fmt.Println(t.processedDocuments)
```
Check out [this commit](https://github.com/operator-framework/combo/pull/38/commits/c066182fbf665d1df3404f491ec2afc16ee8e8ce) for more details!

Thoughts?

pkg/generate/template_test.go
tylerslaton: We do for sure, added a test case for this.

pkg/generate/generate_test.go
tylerslaton: For sure, I went ahead and added an error path check to the `TestBuildTemplate` function. Mind giving that look in [this commit](https://github.com/operator-framework/combo/pull/38/commits/6f59f42ed81e0592e459ae0ce3a809c06fec11f8)?

cmd/eval.go
```diff
@@ -64,7 +73,7 @@ Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
 				return fmt.Errorf("failed to access replacements flag: %w", err)
 			}
 
-			file, err := os.ReadFile(args[0])
+			file, err := os.Open(args[0])
```
perdasilva, line 76: minor and very optional: sometimes it's useful to assign indices as constants. E.g. `const templatePathIndex = 0` then when you do file, err := os.Open(args[templatePathIndex]) you know what it is retrieving from the args. You could also just put it in a var, e.g. `templatePlat := args[0]` or some such. Just a very minor suggestion.

perdasilva: suggestion: maybe also the `file` variable name is too generic? `comboTemplateFile` or something like that could be better?

pkg/generate/generate.go
```diff
@@ -0,0 +1,57 @@
...
+		return nil, fmt.Errorf("failed to build template: %w", err)
+	}
+
+	return &generatorImp{
+		template:     compiledTemplate,
+		combinations: combinations,
+	}, nil
+}
+
+// Evaluate uses specified template and combination stream to build/return the combinations of
```
perdasilva, line 36: Doc string still references the old method name (`Evaluate`)

pkg/generate/generate_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"strings"
+	"testing"
+
+	"github.com/operator-framework/combo/pkg/combination"
+	testdata "github.com/operator-framework/combo/test/assets/generator"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEvaluate(t *testing.T) {
```
perdasilva, line 16: I think this needs to be renamed to `TestGenerate`

pkg/generate/generate_test.go
```diff
@@ -0,0 +1,84 @@
...
+	for _, tt := range []struct {
+		name         string
+		file         io.Reader
+		combinations combination.Stream
+		expected     []string
+		err          error
+	}{
+		{
+			name:     "can process a template",
+			file:     strings.NewReader(testdata.EvaluateInput),
```
perdasilva, line 26: Same here `testdata.EvaluateInput` -> `testdata.GenerateInput` and same for `testdata.EvaluateOutput`

cmd/eval.go
```diff
@@ -78,15 +87,22 @@ Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
 				combination.WithSolveAhead(),
 			)
 
+			generator, err := generate.NewGenerator(file, combinations)
```
perdasilva, line 90: suggestion: Maybe we should either rename the `generate` package, the `Generator` struct (or both)? As it is, it doen't tell us much about _what_ is being generated.

cmd/eval.go
tylerslaton: Nice yeah, these are clean suggestions. Let me throw up an implementation.

cmd/eval.go
tylerslaton: What would you think about renaming each to be `template` and `Builder` respectively? So it would end up looking like this:

```go
import (
    "fmt"
    "github.com/operator-framework/combo/pkg/template"
)

// (...)

templateBuilder, err := template.NewBuilder(templateFile, combinations)
if err != nil {
    return fmt.Errorf("failed to construct generator: %w", err)
}

ctx, cancel := context.WithCancel(context.Background())
defer cancel()

combinedTemplateDocuments, err := templateBuilder.Build(ctx)
if err != nil {
	return fmt.Errorf("failed to generate documents with combinations: %w", err)
}
```

perdasilva (approved):
lgtm - nice work!

cmd/eval.go
```diff
@@ -35,14 +39,20 @@ func formatReplacements(replacements map[string]string) map[string][]string {
...
-func validateFile(file []byte) error {
-	var holder interface{}
-	err := yaml.Unmarshal(file, &holder)
-	if holder == nil {
+// validateFile is a simple wrapper to ensure the file we're using exists, is readable,
+// and is valid YAML
+func validateFile(file io.Reader) error {
+	fileBytes, err := ioutil.ReadAll(file)
+	if err != nil {
+		return ErrCouldNotReadFile
```
awgreene, line 47: I am of the opinion that you should surface the error returned by the `ioutil.ReadAll` function, it could contain  critical debugging data. With this custom error, I only know that I `could not read file`. This error is also [declared in a separate package](https://github.com/operator-framework/combo/pull/38/files#diff-4c6785786c3eaa4a21b69e4a2a47830bb533b95d63b5b6c5bf69bc7e2d76625cR13), and when it is used [the error is surfaced](https://github.com/operator-framework/combo/pull/38/files#diff-4c6785786c3eaa4a21b69e4a2a47830bb533b95d63b5b6c5bf69bc7e2d76625cR27).

cmd/eval.go
```diff
@@ -78,19 +88,26 @@ Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
...
+			}
+
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
 
-			generatedFile, err := generator.Evaluate(ctx, string(file), combinations)
+			combinedTemplateDocuments, err := templateBuilder.Build(ctx)
 			if err != nil {
-				return fmt.Errorf("failed to generate combinations: %w", err)
+				return fmt.Errorf("failed to build documents with combinations: %w", err)
```
awgreene, line 101:
```suggestion
				return fmt.Errorf("failed to build manifests with combinations: %w", err)
```

cmd/eval.go
```diff
@@ -78,19 +88,26 @@ Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
...
+			templateBuilder, err := template.NewBuilder(templateFile, combinations)
+			if err != nil {
+				return fmt.Errorf("failed to construct builder: %w", err)
+			}
+
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
 
-			generatedFile, err := generator.Evaluate(ctx, string(file), combinations)
+			combinedTemplateDocuments, err := templateBuilder.Build(ctx)
```
awgreene, line 99: The Operator Framework typically refers to collections of yaml as manifests and we should do our best to remain consistent. Applies in all places.
```suggestion
			combinedTemplateManifests, err := templateBuilder.Build(ctx)
```

cmd/eval.go
```diff
@@ -78,19 +88,26 @@ Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
...
 			if err != nil {
-				return fmt.Errorf("failed to generate combinations: %w", err)
+				return fmt.Errorf("failed to build documents with combinations: %w", err)
 			}
 
-			if err := validateFile([]byte(generatedFile)); err != nil {
-				return fmt.Errorf("failed to validate file generated: %w", err)
+			combinedTemplate := "---\n" + strings.Join(combinedTemplateDocuments, "\n---\n")
```
awgreene, line 104: [There is no standard convention for newlines across all operating systems](https://stackoverflow.com/questions/3720612/what-is-the-universal-newline-for-all-operating-systems-lf-and-cr), will this cause any issues when the binary is built/ran on different operating systems? We should probably specify which operator systems we support.

This comment does not apply for the combo controller as we control the operating system when building the image.

pkg/template/builder.go
```diff
@@ -0,0 +1,57 @@
...
+	Builder interface {
+		Build(ctx context.Context) ([]string, error)
+	}
+
+	CombinationStream interface {
+		Next(ctx context.Context) (map[string]string, error)
+	}
+)
+
+type builderImp struct {
```
awgreene, line 19:
```suggestion
type builder struct {
```

pkg/template/template.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+var (
+	ErrCouldNotReadFile = errors.New("could not read file")
+)
+
+// template contains an array of documents that can be
+// interacted with with its various functions.
+type template struct {
+	documents          []string
```
awgreene, line 19:
```suggestion
	manifests          []string
```

awgreene (changes requested):
Nice work on this, let's add some e2e tests as well!

cmd/eval.go
tylerslaton: This could probably be handled in the README. It would be a good idea to make an issue where we test the binary on various operating system and see how it handles them. It may also be possible to do that in CI to make sure we don't regress across platforms. Thoughts?

awgreene: SGTM.

pkg/template/template.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func newTemplate(file io.Reader) (template, error) {
+	// Separate the manifests by the yaml separator and build a template with them
+	fileBytes, err := ioutil.ReadAll(file)
+	if err != nil {
+		return template{}, fmt.Errorf("%w: %s", ErrCouldNotReadFile, err.Error())
+	}
+
+	docs := strings.Split(string(fileBytes), "---")
```
awgreene, line 30:
```suggestion
	manifests := strings.Split(string(fileBytes), "---")
```

pkg/template/template.go
```diff
@@ -0,0 +1,68 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"regexp"
+	"strings"
+)
+
+var (
+	ErrCouldNotReadFile = errors.New("could not read file")
```
awgreene, line 13: Is this worth duplicating [here](https://github.com/operator-framework/combo/pull/38/files#diff-56eae6d32e732794aa20bcb63594976b8c9780b628767d3da1e8f85843120fbfR20)?

awgreene (commented):
Looks better, still needs e2e tests.

pkg/template/template.go
tylerslaton: Fixed this up!

pkg/template/template.go
tylerslaton: I made a new `comboErrors` package for errors that are shared across Combo.

pkg/errors/errors.go
```diff
@@ -0,0 +1,8 @@
+package comboerrors
```
awgreene, line 1: Non blocking nit: This should just be `errors`. One can derrive that it contains a list of errors related to the combo project from the fact that the package exists in the combo repository.

awgreene (approved):
Nice work @tylerslaton!

PR #60, opened 2021-11-30
test(e2e): add kind and minikube tests

tylerslaton:
# Summary
Adding small e2e tests to start for kind and minikube. More incoming in future PR's but they're being separated for cognitive load.

# Additions
- e2e action
  - kind and minikube jobs
- `make run-local` target


.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,45 @@
...
+        sudo apt-get install conntrack
+        curl -sLo minikube "$(curl -sL https://api.github.com/repos/kubernetes/minikube/releases/latest | jq -r '[.assets[] | select(.name == "minikube-linux-amd64")] | first | .browser_download_url')"
+        chmod +x minikube
+        sudo mv minikube /bin/
+    - name: Setup minikube
+      run: |
+        minikube config set vm-driver docker
+        minikube start
+    - name: Run Combo on minikube
+      run: make run-local RUN_LOCAL_TYPE=MINIKUBE
```
timflannagan, line 45: I personally think it's overkill to run e2e tests on both kind and minikube and leads to longer e2e times. If we're interested in ensuring compatibility with minikube, maybe make a dedicated action that runs when PRs land in master?

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,45 @@
+name: e2e
+on:
```
timflannagan, line 2: Add a `workflow_dispatch:` trigger as well?

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,45 @@
...
+  pull_request:
+    paths-ignore:
+    - '**/*.md'
+  schedule:
+  - cron: '0 0 * * *' # daily to pick up releases
+jobs:
+  kind:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
```
timflannagan, line 12: I'd downgrade this to the v1 version as there's some problems with v2 checking out the wrong commits.

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,45 @@
...
+      run: |
+        curl -sLo kind "$(curl -sL https://api.github.com/repos/kubernetes-sigs/kind/releases/latest | jq -r '[.assets[] | select(.name == "kind-linux-amd64")] | first | .browser_download_url')"
+        chmod +x kind
+        sudo mv kind /bin/
+    - name: Create kind cluster
+      run: |
+        kind create cluster
+        kind export kubeconfig
+    - name: Run Combo on kind
+      run: make run-local
```
timflannagan, line 26: I'd shift towards using a dedicated `e2e` Makefile target here.

.github/workflows/e2e.yaml
tylerslaton: Think we should downgrade the other actions as well?

timflannagan: Yeah, that would be my preference, but let's do that in a separate PR.

.github/workflows/e2e.yaml
tylerslaton: So would the `minikube` test only run when we merge into main, in this case?

timflannagan: Yeah that's what I had in mind, but I also think you could just drop minikube support entirely and only test on kind clusters, especially since this project is in greenfield stages.

tylerslaton: Let's go with that for now, I think we will eventually want to guarantee Minikube functionality though - right?

tylerslaton:
Closing to bundle in with an upcoming PR.

PR #64, opened 2022-01-05
refactor: address oustanding comments on PR #61

tylerslaton:
# Summary
Fixing a number of issues left over on PR #61 

# Fixes
* Pin specific kind version for e2e action
* Address various small code structure changes
* Update Makefile to be more idiomatic
* Remove recursive Makefile calls
* Update test-e2e and the action that calls it to use Ginkgo
* Move `kubeclient.Update()` call into Eventually handler

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -16,12 +16,16 @@ jobs:
       with:
         go-version: '~1.17'
     - name: Install kind
+      env:
+        KIND_VERSION: v0.11.1
+        KIND_OS: kind-linux-amd64
       run: |
-        curl -sLo kind "$(curl -sL https://api.github.com/repos/kubernetes-sigs/kind/releases/latest | jq -r '[.assets[] | select(.name == "kind-linux-amd64")] | first | .browser_download_url')"
+        curl -sLo kind "https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/${KIND_OS}"
         chmod +x kind
         sudo mv kind /bin/
     - name: Create kind cluster
       run: |
+        kind version
         kind create cluster
         kind export kubeconfig
     - name: Run e2e tests
--- a/Makefile
+++ b/Makefile
@@ -6,9 +6,11 @@ PKG := $(ORG)/combo
 VERSION_PATH := $(PKG)/pkg/version
 GIT_COMMIT := $(shell git rev-parse HEAD)
 DEFAULT_VERSION := v0.0.1
-CONTROLLER_GEN=$(Q)go run sigs.k8s.io/controller-tools/cmd/controller-gen
+CONTROLLER_GEN := $(Q)go run sigs.k8s.io/controller-tools/cmd/controller-gen
 GO_BUILD := $(Q)go build
-PKGS = $(shell go list ./...)
+PKGS := $(shell go list ./...)
+COMBO_VERSION := $(shell git describe || echo $(DEFAULT_VERSION))
+
 
 # Binary build options
 KUBERNETES_VERSION=v0.22.2
@@ -62,22 +64,22 @@ lint: ## Run golangci-lint
 verify: tidy generate format lint ## Verify the current code generation and lint
 	git diff --exit-code
 
-COMBO_VERSION=$(shell git describe || echo $(DEFAULT_VERSION))
-BUILD_FLAGS=CGO_ENABLED=0
 VERSION_FLAGS=-ldflags "-X $(VERSION_PATH).GitCommit=$(GIT_COMMIT) -X $(VERSION_PATH).ComboVersion=$(COMBO_VERSION) -X $(VERSION_PATH).KubernetesVersion=$(KUBERNETES_VERSION)"
-build-cli: ## Build the CLI binary. Speciy VERSION_PATH, GIT_COMMIT, or KUBERNETES_VERSION to change the binary version.
-	$(Q) $(BUILD_FLAGS) go build $(VERSION_FLAGS) -o ./bin/combo
+build-cli: ## Build the CLI binary. Specify VERSION_PATH, GIT_COMMIT, or KUBERNETES_VERSION to change the binary version. You may also specify BUILD_OS and BUILD_ARCH to change the build's binary. 
+	$(Q) CGO_ENABLED=0 GOOS=$(BUILD_OS) GOARCH=$(BUILD_ARCH) go build $(VERSION_FLAGS) -o ./bin/combo
 
 build-container: ## Build the Combo container. Accepts IMAGE_REPO and IMAGE_TAG overrides.
 	docker build . -f Dockerfile -t $(IMAGE)
 
+build-local-container: BUILD_OS=linux
+build-local-container: BUILD_ARCH=amd64
 build-local-container: build-cli ## Build the Combo container from the Dockerfile.local to speed compile time up. Accepts IMAGE_REPO and IMAGE_TAG overrides.
 	docker build . -f Dockerfile.local -t $(IMAGE)
 
 ################
 # Test Targets #
 ################
-.PHONY: test test-unit test-local test-e2e
+.PHONY: test test-unit test-e2e
 
 test: test-unit test-e2e ## Run both the unit and e2e tests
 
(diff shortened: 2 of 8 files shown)


.github/workflows/e2e.yaml
```diff
@@ -17,12 +17,17 @@ jobs:
...
         sudo mv kind /bin/
     - name: Create kind cluster
       run: |
+        kind version
         kind create cluster
         kind export kubeconfig
+    - name: Install Ginkgo
+      run: |
+        go get github.com/onsi/ginkgo/ginkgo@v1.16.4
+        go install github.com/onsi/ginkgo/ginkgo@v1.16.4
```
timflannagan, line 31: Should we add these to tools.go instead?

.github/workflows/e2e.yaml
```diff
@@ -17,12 +17,17 @@ jobs:
         go-version: '~1.17'
     - name: Install kind
       run: |
-        curl -sLo kind "$(curl -sL https://api.github.com/repos/kubernetes-sigs/kind/releases/latest | jq -r '[.assets[] | select(.name == "kind-linux-amd64")] | first | .browser_download_url')"
+        curl -sLo kind "$(curl -sL https://api.github.com/repos/kubernetes-sigs/kind/releases/43733955 | jq -r '[.assets[] | select(.name == "kind-linux-amd64")] | first | .browser_download_url')"
```
timflannagan, line 20: Can we avoid using a magic number (commit? psuedotag?) and instead use a environment variable to track the kind version we want to install:

```
- name: Install kind
  env:
    KIND_VERSION: v0.11.1
  run: |
    curl -sLo kind "$(curl -sL https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION} | jq ...)"
```

.github/workflows/e2e.yaml
tylerslaton: Yeah, that does make more sense now that you point it out.

Makefile
```diff
@@ -62,16 +63,15 @@ lint: ## Run golangci-lint
...
 build-cli: ## Build the CLI binary. Speciy VERSION_PATH, GIT_COMMIT, or KUBERNETES_VERSION to change the binary version.
-	$(Q) $(BUILD_FLAGS) go build $(VERSION_FLAGS) -o ./bin/combo
+	$(Q) CGO_ENABLED=0 go build $(VERSION_FLAGS) -o ./bin/combo
 
 build-container: ## Build the Combo container. Accepts IMAGE_REPO and IMAGE_TAG overrides.
 	docker build . -f Dockerfile -t $(IMAGE)
 
-build-local-container: build-cli ## Build the Combo container from the Dockerfile.local to speed compile time up. Accepts IMAGE_REPO and IMAGE_TAG overrides.
+build-local-container: ## Build the Combo container from the Dockerfile.local to speed compile time up. Accepts IMAGE_REPO and IMAGE_TAG overrides.
+	$(Q) CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build $(VERSION_FLAGS) -o ./bin/combo
```
timflannagan, line 74: It looks like we can just use the `build-cli` target as a prerequisite to this target to avoid skew between dev and CI builds.

Makefile
```diff
@@ -86,29 +86,29 @@ test-unit: ## Run the unit tests
 	$(Q)go test -count=1 -short $(UNIT_TEST_DIRS)
 
 test-e2e: ## Run the e2e tests
-	$(Q)go test -count=1 ./test/e2e
+	ginkgo test ./test/e2e
```
timflannagan, line 89:
```suggestion
	ginkgo run ./test/e2e
```

Makefile
```diff
@@ -86,29 +86,29 @@ test-unit: ## Run the unit tests
...
+
+IMAGE_LOAD_COMMAND=kind load docker-image
+load-image: ## Load-image loads the currently constructed image onto the cluster
+	$(IMAGE_LOAD_COMMAND) $(IMAGE)
+
 deploy: generate ## Deploy the Combo operator to the current cluster
 	kubectl apply --recursive -f manifests
 
+deploy-local: generate ## Deploy the Combo operator to the current cluster
+	kubectl apply --recursive -f manifests
```
timflannagan, line 104: What's the difference between the deploy and deploy-local targets?

tylerslaton: Looks like some left over debugging statement, remove it now.

Makefile
tylerslaton: What do you think of this approach?

```make
VERSION_FLAGS=-ldflags "-X $(VERSION_PATH).GitCommit=$(GIT_COMMIT) -X $(VERSION_PATH).ComboVersion=$(COMBO_VERSION) -X $(VERSION_PATH).KubernetesVersion=$(KUBERNETES_VERSION)"
build-cli: ## Build the CLI binary. Specify VERSION_PATH, GIT_COMMIT, or KUBERNETES_VERSION to change the binary version. You may also specify BUILD_OS and BUILD_ARCH to change the build's binary. 
	$(Q) CGO_ENABLED=0 GOOS=$(BUILD_OS) GOARCH=$(BUILD_ARCH) go build $(VERSION_FLAGS) -o ./bin/combo

build-container: ## Build the Combo container. Accepts IMAGE_REPO and IMAGE_TAG overrides.
	docker build . -f Dockerfile -t $(IMAGE)

build-local-container: BUILD_OS=linux
build-local-container: BUILD_ARCH=amd64
build-local-container: build-cli ## Build the Combo container from the Dockerfile.local to speed compile time up. Accepts IMAGE_REPO and IMAGE_TAG overrides.
	docker build . -f Dockerfile.local -t $(IMAGE)
```

awgreene (approved):
Nice work @tylerslaton, approved.

.github/workflows/e2e.yaml
```diff
@@ -16,13 +16,19 @@ jobs:
...
+        curl -sLo kind "https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/${KIND_OS}"
         chmod +x kind
         sudo mv kind /bin/
     - name: Create kind cluster
       run: |
+        kind version
         kind create cluster
         kind export kubeconfig
+    - name: Install Ginkgo
+      run: go install github.com/onsi/ginkgo/ginkgo@v1.16.4
```
timflannagan, line 32: It looks like this got added to the tools.go, so this should be necessary anymore?

Makefile
```diff
@@ -1,14 +1,15 @@
 ###########################
 # Configuration Variables #
 ###########################
-ORG := github.com/operator-framework
-PKG := $(ORG)/combo
+PKG := $(shell head -n 1 go.mod | awk '{print $$2}')
```
timflannagan, line 4: I don't think it's worth parsing the go.mod to pull out the repository package. If we had to change the repository name, it's not the end of the world having to change any hardcoded assumptions throughout the codebase in multiple places.

Makefile
```diff
@@ -1,14 +1,15 @@
...
 VERSION_PATH := $(PKG)/pkg/version
 GIT_COMMIT := $(shell git rev-parse HEAD)
 DEFAULT_VERSION := v0.0.1
-CONTROLLER_GEN=$(Q)go run sigs.k8s.io/controller-tools/cmd/controller-gen
+CONTROLLER_GEN := $(Q)go run sigs.k8s.io/controller-tools/cmd/controller-gen
 GO_BUILD := $(Q)go build
-PKGS = $(shell go list ./...)
+PKGS := $(shell go list ./...)
+COMBO_VERSION := $(shell git describe || echo $(DEFAULT_VERSION))
```
timflannagan, line 11: I think `git describe` output would be sufficient without falling back to another variable that increases maintenance over time. We don't have to tackle fixing this in this PR though.

test/e2e/controller_test.go
```diff
@@ -137,19 +140,19 @@ var _ = Describe("Combination controller", func() {
...
 		})
 
-		It("should have the combo.ReferencedTemplate label correctly applied on the combination", func() {
+		It("should have the ReferencedTemplate label correctly applied on the combination", func() {
 			Eventually(func(g Gomega) error {
 				var retrievedCombination v1alpha1.Combination
 				err := kubeclient.Get(ctx, types.NamespacedName{Name: validCombinationCRCopy.Name}, &retrievedCombination)
 
-				g.Expect(retrievedCombination.Labels).To(HaveKeyWithValue("combo.ReferencedTemplate", validTemplateCRCopy.Name))
+				g.Expect(retrievedCombination.Labels).To(HaveKeyWithValue(controller.ReferencedTemplateLabel, validTemplateCRCopy.Name))
```
timflannagan, line 152: nit: I'd personally like to avoid mixing gomega (or testify) assertions within an Eventually block, and just handle the error (or whatever check we're interested in performing) directly using conditional checks and appropriately returning.

tools.go
```diff
@@ -5,5 +5,6 @@ package tools
 
 import (
 	_ "github.com/golangci/golangci-lint/cmd/golangci-lint" // Better linting
-	_ "sigs.k8s.io/controller-tools/cmd/controller-gen"     // Generate deepcopy, conversion, and CRDs
+	_ "github.com/onsi/ginkgo/ginkgo@v1.16.4"
```
timflannagan, line 8:
```suggestion
	_ "github.com/onsi/ginkgo/ginkgo"
```

timflannagan (commented):
I think this is largely read to go - only blocking comment is around the ginkgo entry in tools.go.

.github/workflows/e2e.yaml
tylerslaton: Just tested this now and it this line appears necessary to install the Ginkgo CLI used in `make test-e2e`.

timflannagan: I think that's because we can't use the path@version syntax in tools.go like we can with go get/go install, so we need to remove any direct versioning through tools.go, and instead handle any version requirements through the root go.mod.

Makefile
```diff
@@ -86,29 +87,26 @@ test-unit: ## Run the unit tests
 	$(Q)go test -count=1 -short $(UNIT_TEST_DIRS)
 
 test-e2e: ## Run the e2e tests
-	$(Q)go test -count=1 ./test/e2e
```
timflannagan, line 89: I realize this probably needs to be something like `go run github.com/onsi/ginkgo/ginkgo ...` or using a $(GINKGO) variable that runs that command, and updating it here to use the variable.

tylerslaton: Ah, yep that was the missing piece!

timflannagan (approved):

Makefile
tylerslaton: @njhale do you have an opinion on this particular piece?

Makefile
tylerslaton: Let's get this PR through and talk about what we want in a release. Once we get the first release/tag out I'm comfortable removing this line but without it our combo version breaks

PR #66, opened 2022-01-07
refactor(controller): remove ReferencedTemplate label to be more idiomatic

tylerslaton:
# Summary
As was suggested in a recent demo of Combo, a controller adding labels to resources that it did not create is not entirely idiomatic. The original intent behind this change was to optimize the number of iterations that needed to occur when finding combinations relevant to a Template; However, this now seems to not be of much use.

As a result, this PR removes the code that adds and tests for the label.

### DIFF
--- a/pkg/controller/combo_controller.go
+++ b/pkg/controller/combo_controller.go
@@ -9,10 +9,8 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/operator-framework/combo/api/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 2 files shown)


timflannagan (approved):

pkg/controller/combo_controller.go
```diff
@@ -57,32 +55,24 @@ func (c *combinationController) mapTemplateToCombinations(template client.Object
...
-				Name:      combination.Name,
-				Namespace: template.GetNamespace(),
-			},
-		})
+		if combination.Spec.Template == templateName {
+			c.log.Info(fmt.Sprintf("enqueueing %s combination in response to associated %s template being updated", combination.Name, templateName))
+			requests = append(requests, reconcile.Request{
+				NamespacedName: types.NamespacedName{
+					Name:      combination.Name,
+					Namespace: template.GetNamespace(),
```
awgreene, line 72: Non-blocking nit: This is always `""`.

awgreene (approved):
Great work and nice turnaround!
/approve

PR #68, opened 2022-01-07
refactor(makefile): remove DEFAULT_VERSION variable

tylerslaton:


awgreene (approved):
Nice work!

timflannagan (approved):
Nice cleanup - I think we can drop the `--match` here with this repository over time.

PR #70, opened 2022-01-24
feat: added implementation of iterative Next() in combination.go issue#16

### LINKED ISSUE #16: add iterative processing of combinations
# Story
As a user of Combo,
I want to have an iterative approach to combination generation,
So that combinations are generated in a memory efficient way.

# Summary
Currently the `combination` package has the `Stream` interface which defines a `Next()` function. That `Next()` function is currently a stub for the actual functionality we want there. Right now, it processes all the combinations and then iterates over them while cleaning up used combinations along the way to alleviate memory constraints. However, it would be preferable if, with each call to `Next()`, we generate the next combination in the `Stream()` so only one combination is ever in memory at a given time. 

This would help to optimize any large amount of combinations that are generated. The reason this is important is because combination generation has `O(n!)` time complexity but does not necessarily need to have `O(n!)` space complexity (which the algorithm implemented currently does). By implementing the `Next()` function correctly, the space complexity can be greatly reduced down.

# Acceptance Criteria
- Correct implementation of `Next()` function generating combinations iteratively
- Passing tests

jbuchananr:
## Summary
Added correct implementation of the `Next()` function which includes 

- Processing all combinations iteratively on call of `nextIterative()`
- Only creates one combination at a time not using any previous memory
- Created `-p` and `--presolve` flag to specify if combinations are generated all at once or iteratively 
- Converted `All()` to `nextPreSolvedCombination()` 
- Updated combination tests to test implementation of `Next()` with calling both `nextIterativeCombination()` and `nextPreSolvedCombination()`

Closes #16


pkg/combination/combination.go
```diff
@@ -58,28 +75,76 @@ func WithSolveAhead() StreamOption {
 }
 
 // Next gets the next combination from the stream.
-// TODO: Currently this does not solve each combination iteratively and will
-//       need to do so in the future to ensure an optimal use of memory. This
-//       is currently more of a stub to allow consumer packages to maintain its
-// 		 interface.
+// By using this, the stream will solve each combination
+// iteratively.
```
timflannagan, line 79: @tylerslaton Should this be opt-in behavior? Should we expose some configuration knobs such that a user can specify how they'd like to have stream combinations be generated?

pkg/combination/combination.go
```diff
@@ -37,6 +49,11 @@ func NewStream(options ...StreamOption) Stream {
 	for _, option := range options {
 		option(cs)
 	}
+	for key := range cs.args {
+		cs.parameterListFromArgs = append(cs.parameterListFromArgs, key)
+		cs.positionsMapInArgs = append(cs.positionsMapInArgs, 0)
+	}
+	cs.lastUpdatedParameter = len(cs.parameterListFromArgs) - 2
```
timflannagan, line 56: What's the significance of the `2` here? Is this a potential [magic number](https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad#:~:text=In%20programming%2C%20a%20%22magic%20number,in%20more%20than%20one%20place.)?

timflannagan (commented):
Nice start - I had a second to review this before switching to other work. I had a couple of quick questions and style comments, but I didn't focus on the implementation details. Can we add some tests if we're introducing new behavior under-the-hood?

pkg/combination/combination.go
```diff
@@ -58,28 +75,76 @@ func WithSolveAhead() StreamOption {
...
-//       is currently more of a stub to allow consumer packages to maintain its
-// 		 interface.
+// By using this, the stream will solve each combination
+// iteratively.
 func (cs *stream) Next(ctx context.Context) (map[string]string, error) {
-	if cs.solveAhead && !cs.solved {
-		if err := cs.solve(); err != nil {
-			return nil, err
+	// Check to see if anymore combinations exist
+	if !cs.solved {
```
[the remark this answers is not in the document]
jbuchananr: Just added this before seeing this comment!

pkg/combination/combination.go
```diff
@@ -22,11 +22,23 @@ type Stream interface {
...
 }
+
+// args: the raw data from the stream
+// parameterListFromArgs: a list of the names of the parameters
+//                        taken from the stream(args).
+// positionMapInArgs: array of integers that tracks the position
+//                    of the next combination within args.
+// lastUpdatedParameter: an integer that represents the parameter
+//                       we are currently looking at. Intializes as
+//                       second to last value in parameters.
```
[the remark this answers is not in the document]
jbuchananr: I will drop them. Originally, added just to clarify the meaning of the struct variables.

pkg/combination/combination.go
jbuchananr: The `len(cs.parameterListFromArgs) - 2` means we are setting lastUpdatedParameter to the second to last variable in the list of parameterListFromArgs. We do not want to look at the last variable in the list, in my logic for` i = len(cs.positionsMapInArgs) - 1; i > cs.lastUpdatedParameter; i--` you can it goes too `>`. If it would be more intiutive, I think we could change it to `len(cs.parameterListFromArgs) - 1` and make changes elsewhere. It just made more sense when I was writing it to start it at the second to last value in the list.

pkg/combination/combination.go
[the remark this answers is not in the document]
jbuchananr: Not entirely sure how to respond directly to your original comment. But in my latest commit, I also added a new test that I used to check the implementation. This test is more complex and has more combinations.

pkg/combination/combination.go
tylerslaton: Yeah, good call. I think it is good to define this with the `solveAhead` property.

tylerslaton: @jbuchananr to do this you could break out the `Next()` into two smaller private functions.

`nextIterative()` What the old function was
`nextAll()` - What your current function is

Then you would just call one or the other depending on the value of `cs.solveAhead`

pkg/combination/combination.go
timflannagan: I haven't dived too much into the implementation, but I think my preference (generally speaking) is to use a constant variable or adding a comment as to why we're performing that operation. That way, the code is a bit easier to read without having to take a second trying to reason about why a number if being used.

pkg/combination/combination.go
```diff
@@ -58,28 +75,69 @@ func WithSolveAhead() StreamOption {
...
+		// Then updates parameter, and checks to see if the combination is solved.
+		if cs.positionsMapInArgs[cs.lastUpdatedParameter]+1 == len(cs.args[cs.parameterListFromArgs[cs.lastUpdatedParameter]]) {
+			cs.positionsMapInArgs[cs.lastUpdatedParameter] = 0
+			cs.lastUpdatedParameter--
+		}
+		// If combination is not solved, we find the next lastUpdatedParameter,
+		// if lastUpdatedParameter has only 1 argument. Also, will mark as solved
+		// if reach end up parameters.
+		runner := true
+		for (!cs.solved) && (runner) && (cs.lastUpdatedParameter != -1) {
```
tylerslaton, line 128: nit: Are the parenthesis needed here or are they to add clarity?

pkg/combination/combination.go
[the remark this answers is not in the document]
tylerslaton: You could move them to be inline with each property maybe?

pkg/combination/combination.go
jbuchananr: @timflannagan I will go ahead and add a comment. I also think I will go ahead a rename the variable to `nextParameterToUpdate` as this will be more intuitive on why we are looking at the second to last item and not the actual last item.

pkg/combination/combination.go
jbuchananr: Just for clarity, but aren't needed

pkg/combination/combination.go
```diff
@@ -51,39 +60,83 @@ func WithArgs(args map[string][]string) StreamOption {
...
+	}
+	return comboList, nil
 }
 
-// All retrieves all of the combinations from the stream
-func (cs *stream) All() ([]map[string]string, error) {
+// nextAll generates all combinations at once.
+// Once combinations are generated it will return the next combination.
+func (cs *stream) nextAll() (map[string]string, error) {
+	fmt.Println("nextAll() used.")
```
tylerslaton, line 139: nit: looks like a print statement broke through

pkg/combination/combination.go
```diff
@@ -51,39 +60,83 @@ func WithArgs(args map[string][]string) StreamOption {
...
-func (cs *stream) Next(ctx context.Context) (map[string]string, error) {
-	if cs.solveAhead && !cs.solved {
-		if err := cs.solve(); err != nil {
-			return nil, err
-		}
+// nextIterative gets the next combination from the stream.
+// By using this, the stream will solve each combination
+// iteratively.
+func (cs *stream) nextIterative() (map[string]string, error) {
+	fmt.Println("nextIterative() used.")
```
tylerslaton, line 73: nit: looks like a print statement broke through

pkg/combination/combination_test.go
```diff
@@ -35,18 +35,40 @@ var combinationTests = []struct {
...
+	},
 }
 
 func TestAll(t *testing.T) {
 	for _, tt := range combinationTests {
 		t.Run(tt.name, func(t *testing.T) {
 			combinationStream := NewStream(
 				WithArgs(tt.input),
-				WithSolveAhead(),
+				WithSolveAhead(true),
```
tylerslaton, line 53: Since we're remove the `cs.All()` method, we could rename this function to be `TestNextAll()` and the `TestNext()` that currently exists to be `TestNextIterative`. Though I would prefer that both cases be covered by the same test. I.E, just one single `TestNext` that tests both of the combination solution methods.

pkg/combination/combination.go
```diff
@@ -51,39 +60,83 @@ func WithArgs(args map[string][]string) StreamOption {
...
+		cs.solved = true
+	}
+	return comboList, nil
 }
 
-// All retrieves all of the combinations from the stream
-func (cs *stream) All() ([]map[string]string, error) {
+// nextAll generates all combinations at once.
+// Once combinations are generated it will return the next combination.
+func (cs *stream) nextAll() (map[string]string, error) {
```
tylerslaton, line 138: The more I think about it, the more I realize that `nextAll` is somewhat ambiguous. Maybe it would make sense to say something like `nextWithSolveAhead` or something along those lines.

cmd/eval.go
```diff
@@ -43,6 +46,7 @@ func formatReplacements(replacements map[string]string) map[string][]string {
 // and is valid YAML
 func validateFile(file io.Reader) error {
 	fileBytes, err := ioutil.ReadAll(file)
+	fmt.Println(string(fileBytes))
```
tylerslaton, line 49: nit: left over print statement!

cmd/eval.go
```diff
@@ -85,7 +89,7 @@ Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
 
 			combinations := combination.NewStream(
 				combination.WithArgs(formatReplacements(replacements)),
-				combination.WithSolveAhead(),
+				combination.WithSolveAhead(useSolvedAhead),
```
tylerslaton, line 92: This is cleaner in this particular spot, but `WithSolveAhead` becomes a slightly confusing argument now. Originally it was just a boolean toggle but now it takes a boolean. I would be okay with just conditionally setting this value in the `eval` command using the old method or renaming the `WithSolveAhead` option function.

cmd/eval.go
```diff
@@ -19,10 +19,13 @@ import (
 var (
 	ErrEmptyFile      = errors.New("empty file")
 	FilePathArgsIndex = 0
+	useSolvedAhead    bool
 )
 
 func init() {
 	evalCmd.Flags().StringToStringP("replacements", "r", map[string]string{}, "Key value pair of comma delimited values. Example: 'NAMESPACE=foo,bar'")
+	evalCmd.Flags().BoolVarP(&useSolvedAhead, "all", "a", false, "Toggles how combinations are generated. When applied combinations are generated all at once compared to iteratively.")
+	// evalCmd.Flags().StringToStringP
```
tylerslaton, line 28: nit: left over comment

tylerslaton (commented):
Looking good! Some nits, suggestions and clarifications. Let me know if I can clarify at all.

pkg/combination/combination.go
jbuchananr: @tylerslaton Decided to rename the method to `nextPreSolvedCombinations()` as it is more intuitive to what is occurring. I also changed the flag in this case to '-presolve' and '-ps'.

tylerslaton: Oh nice, I like that way more. Would the method make more sense as `nextPreSolvedCombination()` since it is returning a single combination?

pkg/combination/combination_test.go
jbuchananr: Implemented one singular TestNext() that combines the two tests.

cmd/eval.go
```diff
@@ -19,10 +19,12 @@ import (
 var (
 	ErrEmptyFile      = errors.New("empty file")
 	FilePathArgsIndex = 0
+	useSolvedAhead    bool
 )
 
 func init() {
 	evalCmd.Flags().StringToStringP("replacements", "r", map[string]string{}, "Key value pair of comma delimited values. Example: 'NAMESPACE=foo,bar'")
+	evalCmd.Flags().BoolVarP(&useSolvedAhead, "presolve", "p", false, "Toggles how combinations are generated. When applied combinations are generated all at once compared to iteratively.")
```
timflannagan, line 27: Let's avoid using initializing a global variable here, and instead get the value of the "presolve" option in the [`RunE(...)`](https://github.com/operator-framework/combo/blob/74719d2173a762660e7745b00f221bb2f51c52e8/cmd/eval.go#L73) function. You can find an example of this on line 74 where we get the value of the `replacements` CLI flag.

```go
useSolveAhead, err := cmd.Flags().GetBool("presolve")
if err != nil {
  return err
}
...
```

cmd/eval.go
```diff
@@ -83,10 +85,17 @@ Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
...
+			if useSolvedAhead {
+				combinations = combination.NewStream(
+					combination.WithArgs(formatReplacements(replacements)),
+					combination.WithSolveAhead(),
+				)
+			} else {
+				combinations = combination.NewStream(
+					combination.WithArgs(formatReplacements(replacements)),
+				)
+			}
```
timflannagan, line 98: I think this might be a candidate for a slight refactor - we could update that `combination.WithSolveAhead()` StreamOption to take a bool, and that method is responsible for setting the value of the `cs.solveAhead = solveAhead`. @tylerslaton any thoughts? We'd be able to remove the need to conditionally check the `useSolveAhead` variable here, and remove the nesting that reduces readability.

jbuchananr: Another alternative could be instead of using `WithSolveAhead()` we create getter and setter methods such as `setSolveAhead()` and `getSolveAhead()`, so there is no confusion on why `WithSolveAhead()` that is currently used to set solve ahead as `true` takes a variable.

timflannagan: Yeah that's one possibility - I'm generally a big fan of this functional options pattern as it helps maintain API flexibility, but the discoverability of the various options you can use isn't always ideal, especially if those individual options aren't prefixed with `With*(...)`. I'd like to avoid any getter/setters in the API as that feels like an anti-pattern for Go these days.

Anyways, not a huge deal as this project is still in the early dev stages, so we can always re-evaluate our API over time before committing to anything long term.

jbuchananr: For sure, I know I originally had made that change to simplify the code, but Tyler and I ended up deciding it could end up being confusing.

cmd/eval.go
jbuchananr: Got it. I implemented the change, and it will be pushed in the next commit. Waiting to make commit until we can get some feed back on what Tyler thinks about changing `WithSolveAhead` to take a bool.

cmd/eval.go
tylerslaton: Yeah, Jon had originally made a change where `WithSolveAhead()` took a bool and this got cleaned up. I was originally opposed to it, but I think that this does become cleaner to a degree that makes sense now that we're talking about it here.

tylerslaton: Alternatively, we could go with an "optional parameter" approach:

https://stackoverflow.com/questions/19612449/default-value-in-gos-method

jbuchananr: I like the optional approach. 
Here's how it would work inline
```
func WithSolveAhead(boolVar ...bool) StreamOption {
	boolVarValue := false
	if len(boolVar) > 0 {
		boolVarValue = boolVar[0]
	}
	return func(cs *stream) {
		cs.solveAhead = boolVarValue
	}
}
```

tylerslaton: I like that, @timflannagan thoughts?

timflannagan: I think that implementation is introducing extra complexity, but let's just forward on whatever we're comfortable with now. I wouldn't block a PR on this.

tylerslaton: Let's just go with the required boolean (like what Jon had before) and call it a day on this thread then.

cmd/eval.go
```diff
@@ -13,7 +13,7 @@ import (
 	comboErrors "github.com/operator-framework/combo/pkg/errors"
 	"github.com/operator-framework/combo/pkg/template"
 	"github.com/spf13/cobra"
-	"gopkg.in/yaml.v3"
+	"gopkg.in/yaml.v2"
```
tylerslaton, line 16: Oh I just noticed this downgrade, any idea what caused this?

cmd/eval.go
jbuchananr: Got it, I committed with optional, but I will re commit with the required.

cmd/eval.go
jbuchananr: Not entirely sure, I'll change it back and make sure it isn't causing issues. I think it could've been changed when I was playing around with Cobra the other day.

tylerslaton (approved):
Great work!

/lgtm

cmd/eval.go
timflannagan: I'm not sure I entirely agree that `WithSolveAhead(...)` becomes more confusing from an API standpoint. You're essentially just exposing setter-esq functionality as an API option, which is already a fairly prevalent pattern. Maybe we can look into making the API surface more consistent over time if there are any UX problems when interacting with the combinations API?

pkg/combination/combination.go
```diff
@@ -51,39 +58,85 @@ func WithArgs(args map[string][]string) StreamOption {
...
+	// looking for the first updatable value then breaks loop.
+	// Otherwise, resets values to zero, we know to update last parameter based off i.
+	var i int
+	for i = len(cs.positionsMapInArgs) - 1; i > cs.nextParameterToUpdate; i-- {
+		if cs.positionsMapInArgs[i]+1 < len(cs.args[cs.parameterListFromArgs[i]]) {
+			cs.positionsMapInArgs[i]++
+			break
+		} else {
+			cs.positionsMapInArgs[i] = 0
+		}
```
timflannagan, line 106: nit/not blocking: you can reduce the amount of nesting by removing the if/else conditional structure due to the `break` keyword being used:

```go
if cs.positionsMapInArgs[i]+1 < len(cs.args[cs.parameterListFromArgs[i]]) {
	cs.positionsMapInArgs[i]++
	break
}
cs.positionsMapInArgs[i] = 0
```

pkg/combination/combination.go
```diff
@@ -51,39 +58,85 @@ func WithArgs(args map[string][]string) StreamOption {
...
+		// if nextParameterToUpdate has only 1 argument. Also, will mark as solved
+		// if reach end up parameters.
+		runner := true
+		for !cs.solved && runner && cs.nextParameterToUpdate != -1 {
+			if len(cs.args[cs.parameterListFromArgs[cs.nextParameterToUpdate]]) == 1 {
+				cs.nextParameterToUpdate--
+			} else {
+				runner = false
+				cs.positionsMapInArgs[cs.nextParameterToUpdate]++
+			}
```
timflannagan, line 128: nit/not blocking: you can remove the if/else conditional structure here, which improves readability by using the `continue` keyword:

```go
if len(cs.args[cs.parameterListFromArgs[cs.nextParameterToUpdate]]) == 1 {
	cs.nextParameterToUpdate--
	continue
}
runner = false
cs.positionsMapInArgs[cs.nextParameterToUpdate]++
```

pkg/combination/combination_test.go
```diff
@@ -69,7 +90,7 @@ func TestNext(t *testing.T) {
 			for {
 				next, err := combinationStream.Next(ctx)
 				if !errors.Is(err, tt.expected.err) {
-					t.Fatal("error received while processing combination stream:", err)
+					t.Fatal("[TEST nextPreSolvedCombinations()] error received while processing combination stream:", err)
```
timflannagan, line 93: nit/not blocking: Adding the identifying `[Test ...]` shouldn't be necessary here as Go should outline which test case failed.

cmd/eval.go
```diff
@@ -23,6 +23,10 @@ var (
 
 func init() {
 	evalCmd.Flags().StringToStringP("replacements", "r", map[string]string{}, "Key value pair of comma delimited values. Example: 'NAMESPACE=foo,bar'")
+
+	var tempSolvedAhead bool
```
timflannagan, line 27: We can avoid declaring this variable entirely by using the `BoolVar` function instead of `BoolVarP`. The `tempSolvedAhead` variable is only scoped to the init function (and therefore not global), and we don't use it's value anywhere else outside of the `RunE(...)` driver code.

cmd/eval.go
```diff
@@ -74,18 +78,19 @@ Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
 				return fmt.Errorf("failed to access replacements flag: %w", err)
 			}
 
-			templateFile, err := os.Open(args[FilePathArgsIndex])
+			useSolvedAhead, err := cmd.Flags().GetBool("presolve")
```
timflannagan, line 81: :+1:

pkg/combination/combination.go
```diff
@@ -96,8 +149,9 @@ func (cs *stream) All() ([]map[string]string, error) {
 		}
 		return nil, ErrCombinationsNotSolved
 	}
-
-	return cs.combinations, nil
+	combo := cs.combinations[0]
+	cs.combinations = cs.combinations[1:]
```
timflannagan, line 153: Do we need to length check the `cs.combinations` variable in order to avoid attempting to blindly index into the array and producing a panic?

pkg/combination/combination.go
```diff
@@ -51,39 +58,85 @@ func WithArgs(args map[string][]string) StreamOption {
...
+		// Checks to see if this is the last argument of the parameter.
+		// Then updates parameter, and checks to see if the combination is solved.
+		if cs.positionsMapInArgs[cs.nextParameterToUpdate]+1 == len(cs.args[cs.parameterListFromArgs[cs.nextParameterToUpdate]]) {
+			cs.positionsMapInArgs[cs.nextParameterToUpdate] = 0
+			cs.nextParameterToUpdate--
+		}
+		// If combination is not solved, we find the next nextParameterToUpdate,
+		// if nextParameterToUpdate has only 1 argument. Also, will mark as solved
+		// if reach end up parameters.
+		runner := true
```
timflannagan, line 121: nit/not blocking: I think we can rename this variable to help improve readability. Maybe `continueIterating`?

pkg/combination/combination.go
timflannagan: I think we might also be reaching complexity creep with this method now too. I see a lot of mutation around the `cs.solved` field, which is fine for now, but is likely a candidate for refactoring down the line.

pkg/combination/combination_test.go
```diff
@@ -35,25 +35,46 @@ var combinationTests = []struct {
...
 		t.Run(tt.name, func(t *testing.T) {
 			combinationStream := NewStream(
 				WithArgs(tt.input),
-				WithSolveAhead(),
 			)
-			got, err := combinationStream.All()
-			if !errors.Is(err, tt.expected.err) {
-				t.Fatal("error received while retreiving all combinations:", err)
+
+			var got []map[string]string
```
timflannagan, line 56: nit/not blocking: we may want to refactor this test packages setup to use [table-driven testing](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) and add the `got` and a `want` fields to a custom struct. It looks like we also check the types of errors that are produced, so having an expectErr, or wantErr field could be appropriate too.

pkg/combination/combination_test.go
```diff
@@ -35,25 +35,46 @@ var combinationTests = []struct {
...
+				got = append(got, next)
 			}
-			require.ElementsMatch(t, got, tt.expected.combinations, "Combos generated incorrectly")
+			require.ElementsMatch(t, got, tt.expected.combinations, "[TEST nextIterative()] Combos generated incorrectly")
 		})
 	}
-}
 
-func TestNext(t *testing.T) {
+	// Test 2: Uses nextPreSolvedCombinations()
```
timflannagan, line 77: I think my preference would be to avoid having to separate tests within a single top-level test spec. It seems like my suggestion in https://github.com/operator-framework/combo/pull/70/files#r795839125 is more applicable now, and you could also pass a `solveAhead` parameter as a struct field that the testing code can handle, e.g. `WithSolveAhead(tt.solveAhead)`.

pkg/combination/combination_test.go
```diff
@@ -35,25 +35,46 @@ var combinationTests = []struct {
...
 		},
 	},
+	{
+		name:  "standard set of long args",
+		input: testdata.LongCombinationInput,
+		expected: expected{
+			combinations: testdata.LongCombinationOutput,
+			err:          nil,
+		},
+	},
```
timflannagan, line 45: I think we should avoid trying to tack on extra test cases to this global testing spec, and instead define it in-line using table-driven tests.

timflannagan:
I think we're getting close to the finish line. I tried to avoid restructuring the algorithm you already implemented, and mainly focused on getting the code to a cleaner state. My main feedback was we should aim to reduce global state (especially shared) as much as possible, work towards using table-driven tests, and cleanup some variable names. We can make some easy readability wins by reducing nesting as much as possible for reviewers and devs committing to this codebase as it reduces the amount of mental strain you need to track while glossing over a method/function/etc. This can be done by using first-class keywords like `break` and `continue` while iterating, on top of returning early whenever possible, or keeping function scope small in nature.

pkg/combination/combination_test.go
tylerslaton: Yeah, I agree. It made more sense when we were using the same data for two tests but does not anymore. A good example of what this should look like now can be found in the builder test.

https://github.com/operator-framework/combo/blob/main/pkg/template/builder_test.go

timflannagan: Good catch - I meant to link to a concrete example, but it looks like that fell through the cracks. Try and take a look at what Tyler had linked as it should paint a decent picture of the setup we're talking about.

pkg/combination/combination.go
jbuchananr: Nope. The only time this would fail is `len(cs.combinations) == 0` which is checked just before this. If there is only one element originally `cs.combinations[1:]` will just be an empty array.

pkg/combination/combination.go
```diff
@@ -51,39 +58,85 @@ func WithArgs(args map[string][]string) StreamOption {
...
 
-	if len(cs.combinations) == 0 {
-		if cs.solved {
-			return nil, nil
-		}
-		return nil, ErrCombinationsNotSolved
+	// Edge case: No parameterListFromArgs
+	if len(cs.parameterListFromArgs) == 0 {
+		cs.solved = true
+		return nil, ErrNoArgsSet
```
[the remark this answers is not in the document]
jbuchananr: I believe this error is handled when the input is "empty"

timflannagan (approved):
Let's handle refactoring the combinations_test.go in a follow-up PR. I'd like to also see us make some minor cleanup (variable naming, conditional structure, removing unneeded commens, etc.) but I don't want to block this work from merging in the meantime.

PR #73, opened 2022-01-28
fix: Corrected 'eval' function to correctly output combinations for issue #72

### LINKED ISSUE #72: fix eval command to correctly output combinations
## Story
As a user of combo, I want to be able to run the evaluation command and have it output the correct combinations. 
E.g `./bin/combo eval -r TARGET_NAMESPACE=1,2 -r TARGET_GROUP=3,4 test.yaml` -> outputs combinations from test.yaml

## Summary

Currently, when running the `eval` command nothing is outputted. For example, if we run `./bin/combo eval -r TARGET_NAMESPACE=1,2 -r TARGET_GROUP=3,4 test.yaml > output.yaml` with `test.yaml` as sample input we would expect combinations to be outputted to `output.yaml`. However, there is currently no output being generated. 
## Acceptance Criteria 

- Output from eval cmd is generated 
- Passes e2e tests

## Additional 
- Evaluate how to handle empty file inputs

jbuchananr:
## Summary
- fixed test case of builder to correctly validate yaml input
- passes e2e and unit tests
- correctly outputs combinations with the call of `eval` command in the CLI 

Closes #72

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -4,16 +4,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
-	"io/ioutil"
 	"os"
 	"strings"
 
 	"github.com/operator-framework/combo/pkg/combination"
-	comboErrors "github.com/operator-framework/combo/pkg/errors"
 	"github.com/operator-framework/combo/pkg/template"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
-	"gopkg.in/yaml.v3"
 )
 
 var (
@@ -39,22 +36,6 @@ func formatReplacements(replacements map[string]string) map[string][]string {
 		formattedReplacements[key] = strings.Split(val, ",")
 	}
 	return formattedReplacements
-}
-
-// validateFile is a simple wrapper to ensure the file we're using exists, is readable,
-// and is valid YAML
-func validateFile(file io.Reader) error {
-	fileBytes, err := ioutil.ReadAll(file)
-	if err != nil {
-		return fmt.Errorf("%w: %s", comboErrors.ErrCouldNotReadFile, err.Error())
-	}
-
-	if len(fileBytes) == 0 {
-		return ErrEmptyFile
-	}
-
-	var holder interface{}
-	return yaml.Unmarshal(fileBytes, &holder)
 }
 
 var (
@@ -98,17 +79,17 @@ Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
 
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
-
 			combinedTemplateManifests, err := templateBuilder.Build(ctx)
 			if err != nil {
 				return fmt.Errorf("failed to build manifests with combinations: %w", err)
 			}
 
-			combinedTemplate := "---\n" + strings.Join(combinedTemplateManifests, "\n---\n")
-
-			if err := validateFile(strings.NewReader(combinedTemplate)); err != nil {
-				return fmt.Errorf("failed to validate combined template constructed: %w", err)
+			if len(combinedTemplateManifests) == 0 {
+				logrus.Warn("resulting combinations are empty")
+				return nil
 			}
+
+			combinedTemplate := "---\n" + strings.Join(combinedTemplateManifests, "\n---\n")
 
 			fmt.Println(combinedTemplate)
 
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/jinzhu/copier v0.3.2
 	github.com/onsi/ginkgo v1.16.4
 	github.com/onsi/gomega v1.15.0
+	github.com/sirupsen/logrus v1.8.1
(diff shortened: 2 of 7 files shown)


tylerslaton:
Will review this once #70 goes through as the history is out of date currently.

cmd/eval.go
```diff
@@ -54,6 +54,7 @@ func validateFile(file io.Reader) error {
 	}
 
 	var holder interface{}
+
 	return yaml.Unmarshal(fileBytes, &holder)
 }
```
tylerslaton, line 59: I would say that the `validateFile` function should no longer be necessary. We should instead be validating within the `template` package before this is needed. Basically, when we call the `templateBuilder.Build()` I'd say we should just go ahead and call the template's validate command which would cover this. Thoughts?

tylerslaton (changes requested):
Making good progress, have a couple of questions I'd like to see addressed but this is great!

test/assets/template/testdata.go
```diff
@@ -4,91 +4,91 @@ var BuildInput = `---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-	name: deployment-reader
```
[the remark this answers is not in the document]
jbuchananr: Interesting. I was having issues with this with unit tests not passing with this space here. We can look at it tomorrow @ 9:30.

cmd/eval.go
jbuchananr: I am perfectly fine with removing the validation function from `eval.go`. From what I can tell, in the previous iteration of validation, we also were validating the output of `combinedTemplate` which is why the function was left over.

pkg/template/template.go
```diff
@@ -17,13 +23,28 @@ type template struct {
...
+	if len(t.manifests) == 0 {
+		return ErrEmptyFile
+	}
+
+	for _, manifest := range t.manifests {
+		var holder interface{}
+		if err := yaml.Unmarshal([]byte(manifest), &holder); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
jbuchananr: I like this a lot. I will get this change added to the next commit.

cmd/eval.go
```diff
@@ -92,20 +93,21 @@ Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
...
 			defer cancel()
-
 			combinedTemplateManifests, err := templateBuilder.Build(ctx)
 			if err != nil {
 				return fmt.Errorf("failed to build manifests with combinations: %w", err)
 			}
 
 			combinedTemplate := "---\n" + strings.Join(combinedTemplateManifests, "\n---\n")
 
+			// validate created file
```
timflannagan, line 110: nit: I think we can remove this comment as the function name should be self-explanatory

cmd/eval.go
```diff
@@ -92,20 +93,21 @@ Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
 			)
 
 			templateBuilder, err := template.NewBuilder(templateFile, combinations)
+
```
timflannagan, line 96: nit: let's remove this newline

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/onsi/gomega v1.15.0
 	github.com/spf13/cobra v1.2.1
 	github.com/stretchr/testify v1.7.0
-	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
+	gopkg.in/yaml.v2 v2.4.0
```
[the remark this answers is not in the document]
jbuchananr: No idea, I believe it was something when I was learning stuff about flags and the yaml version got changed there on accident, not entirely sure though.

cmd/eval.go
jbuchananr: Got it.

cmd/eval.go
tylerslaton: I think we can honestly get rid of the `combinedTemplate` and the validation function as we are essentially adding explicitly valid YAML to whatever is inside of the template.

cmd/eval.go
```diff
@@ -98,18 +78,13 @@ Example: combo eval -r REPLACE_ME=1,2,3 path/to/file
 
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
-
 			combinedTemplateManifests, err := templateBuilder.Build(ctx)
 			if err != nil {
 				return fmt.Errorf("failed to build manifests with combinations: %w", err)
 			}
 
 			combinedTemplate := "---\n" + strings.Join(combinedTemplateManifests, "\n---\n")
```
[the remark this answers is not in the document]
tylerslaton: Hm, this is interesting. Yeah I'm 100% down to improve how manifests are combined together to be more idiomatic as the current method feels a bit hacky. Maybe it is valuable to create an issue to track/address this?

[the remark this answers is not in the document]
exdx: Sure, I can open a tracking issue

tylerslaton (approved):
Great work here, Jon!

PR #75, opened 2022-02-01
.github: Remove paths_ignore stanza from relevant workflows

timflannagan:
Remove the paths_ignore stanza as it doesn't report back workflow status
correctly when checks are skipped due to markdown-only PRs, which
doesn't play well with how branch protection rules are evaluated.


tylerslaton (approved):
Looks good to me!

PR #78, opened 2022-02-16
feat: apply combinations to the cluster 

### LINKED ISSUE #17: flag to apply evaluation for controller
# Story
As a user of Combo,
I want to be able to apply the evaluation of Combo,
So that there is an extra layer of optional automation.

# Summary 
This relies on #2 being complete. As it currently stands, #2 will make it so that the template and combinations can be evaluated but will not be applied to the cluster. This task should add an optional flag that will enable the creator to have Combo automatically apply the evaluation to the cluster.

# Acceptance Criteria
- Can set a flag to apply the evaluation to the cluster
- Has E2E test(s) to validate the logic and implementation

exdx:
Closes #17 

This PR introduces a new `Apply` boolean field on the Combination Spec that indicates that the finished evaluation should then be applied to the cluster. Note that since the combo operator by default only has RBAC permissions to manage its own APIs, creation of any objects outside of the combo APIs would fail. For exampling, templating out a configmap and then applying it to the cluster will fail with a permissions error. The idea is cluster-admins would manually adjust the permissions granted to combo in order to then be able to apply all evaluations to the cluster. 

Signed-off-by: Daniel Sover <dsover@redhat.com>


api/v1alpha1/combination_types.go
```diff
@@ -27,6 +27,8 @@ const (
 	ReasonTemplateBodyInvalid = "TemplateBodyInvalid"
 	ReasonEvaluationsInvalid  = "EvaluationsInvalid"
 	ReasonProcessed           = "Processed"
+	ReasonCreationFailed      = "CreationFailed"
```
tylerslaton, line 30: nit: I'd say `ApplyFailed`/`ApplySucceeded` makes this a bit clearer since the flag is `Apply`. What are your thoughts on that?

pkg/controller/combo_controller.go
```diff
@@ -153,6 +155,41 @@ func (c *combinationController) Reconcile(ctx context.Context, req ctrl.Request)
 		Message:            "evaluations successfully processed",
 	})
 
+	// If manifest application is indicated, create generated manifests on-cluster.
+	if combination.Spec.Apply{
```
tylerslaton, line 159: Maybe it makes sense to add a `v1alpha.ReasonProcessing` status update call here before the loop starts in case it takes awhile.

pkg/controller/combo_controller.go
```diff
@@ -153,6 +155,41 @@ func (c *combinationController) Reconcile(ctx context.Context, req ctrl.Request)
...
+			}
+			combination.SetStatusCondition(metav1.Condition{
+				Type:               v1alpha1.TypeFinished,
+				Status:             metav1.ConditionTrue,
+				Reason:             v1alpha1.ReasonCreationSucceeded,
+				LastTransitionTime: metav1.NewTime(time.Now()),
+				Message:            fmt.Sprintf("created manifest %s on-cluster", obj.GetName()),
+			})
+		}
+	}
```
tylerslaton, line 191: What are your thoughts on this? This seems to make a bit more sense to me as we'd want to update with a successful reason once all manifests have been processed (since there are few applies happening in that loop):

```suggestion
			if err != nil {
				combination.SetStatusCondition(metav1.Condition{
					Type:               v1alpha1.TypeInvalid,
					Status:             metav1.ConditionFalse,
					Reason:             v1alpha1.ReasonCreationFailed,
					LastTransitionTime: metav1.NewTime(time.Now()),
					Message:            fmt.Sprintf("failed to create manifest %s on-cluster: %s", obj.GetName(), err.Error()),
				})
			}
		}
		combination.SetStatusCondition(metav1.Condition{
				Type:               v1alpha1.TypeFinished,
				Status:             metav1.ConditionTrue,
				Reason:             v1alpha1.ReasonCreationSucceeded,
				LastTransitionTime: metav1.NewTime(time.Now()),
				Message:            fmt.Sprintf("created manifest %s on-cluster", obj.GetName()),
			})
	}
```

tylerslaton (commented):
Took a quick look, overall amazing work. This is nearly there just need to get it cleaned up and then we're good to go.

api/v1alpha1/combination_types.go
exdx: that makes sense 👍

pkg/controller/combo_controller.go
```diff
@@ -153,6 +157,53 @@ func (c *combinationController) Reconcile(ctx context.Context, req ctrl.Request)
...
+				combination.SetStatusCondition(metav1.Condition{
+					Type:               v1alpha1.TypeInvalid,
+					Status:             metav1.ConditionFalse,
+					Reason:             v1alpha1.ReasonApplyFailed,
+					LastTransitionTime: metav1.NewTime(time.Now()),
+					Message:            fmt.Sprintf("failed to decode manifest %s to apply to cluster: %s", obj.GetName(), err.Error()),
+				})
+				return reconcile.Result{}, errors.NewAggregate([]error{err, c.Status().Update(ctx, combination)})
+			}
+			err = c.Create(ctx, obj)
```
timflannagan, line 185: Do we need to support updating resources as well, or just the application of those resources?

test/e2e/controller_test.go
```diff
@@ -218,4 +220,197 @@ var _ = Describe("Combination controller", func() {
...
+				var conditionReasons []string
+				for _, condition := range retrievedCombination.Status.Conditions {
+					conditionReasons = append(conditionReasons, condition.Reason)
+				}
+				g.Expect(conditionReasons).To(ContainElement(v1alpha1.ReasonProcessed))
+				// g.Expect(conditionReasons).To(ContainElement(v1alpha1.ReasonApplySucceeded))
+
+				// fetch created template on-cluster and ensure templating worked as expected
+				var embeddedTemplate v1alpha1.Template
+				err = kubeclient.Get(ctx, types.NamespacedName{Name: "embedded-template-dev"}, &embeddedTemplate)
```
timflannagan, line 407: Handle error first, and then check whether the labels match our expectation? Alternatively, handle the error and updating the Eventually handler to check for the return value of the template labels?

pkg/controller/combo_controller.go
```diff
@@ -153,6 +157,53 @@ func (c *combinationController) Reconcile(ctx context.Context, req ctrl.Request)
 		Message:            "evaluations successfully processed",
 	})
 
+	// If manifest application is indicated, create generated manifests on-cluster.
+	if combination.Spec.Apply {
```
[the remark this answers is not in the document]
tylerslaton: Yeah, I agree with that. I could see something like:

```Go
if combination.Spec.Apply {
    if err := applyManifests(manifests); err != nil {
        // error handling stuffs ...
    }
}
```
Where applyManifests is just all of this logic

timflannagan (commented):
This looks reasonable to me - only major comments were around refactoring to have the business logic housed in it's own method/function and calling that during the Reconcile driver method. I saw a couple of places in the e2e test code where it didn't appear we were handling the error return immediately, but I'd also like to wrap any CRUD-related API call in an eventually to future proof ourselves.

pkg/controller/combo_controller.go
joelanford: I'd suggest `c.Patch` with the server-side-apply option.

joelanford (commented):
At what point do we clean up any previously applied manifests that are no longer rendered by the current state of the template/combo?

pkg/controller/combo_controller.go
```diff
@@ -153,6 +157,53 @@ func (c *combinationController) Reconcile(ctx context.Context, req ctrl.Request)
...
+		// update status to indicate processing
+		combination.SetStatusCondition(metav1.Condition{
+			Type:               v1alpha1.TypeInProgress,
+			Status:             metav1.ConditionTrue,
+			Reason:             v1alpha1.ReasonProcessing,
+			LastTransitionTime: metav1.NewTime(time.Now()),
```
joelanford, line 167: We shouldn't be updating the `LastTransitionTime` unless the `Status` has changed since the last time we updated this condition. This kind of thing can result in a hotloop where you update the condition with this timestamp, and then reconcile because the status was updated and then update the transition time stamp again, and then reconcile again, etc.

joelanford: There are helpers for this in the apimachinery repo: https://github.com/kubernetes/apimachinery/blob/ec22c436956176895311ebeee513ca9dee0b3aa5/pkg/api/meta/conditions.go#L30

joelanford: Upon further inspection, it looks like `combination.SetStatusCondition` just calls that function under the hood. In that case, I think you can just drop all explicit setting of `LastTransitionTime`

pkg/controller/combo_controller.go
```diff
@@ -153,6 +157,53 @@ func (c *combinationController) Reconcile(ctx context.Context, req ctrl.Request)
 		Message:            "evaluations successfully processed",
 	})
 
+	// If manifest application is indicated, create generated manifests on-cluster.
+	if combination.Spec.Apply {
+		// update status to indicate processing
+		combination.SetStatusCondition(metav1.Condition{
+			Type:               v1alpha1.TypeInProgress,
+			Status:             metav1.ConditionTrue,
```
joelanford, line 165: Does this ever get set back to `False` once we're done progressing? If not, it should.

pkg/controller/combo_controller.go
```diff
@@ -153,6 +157,53 @@ func (c *combinationController) Reconcile(ctx context.Context, req ctrl.Request)
...
+		})
+
+		for _, manifest := range generatedManifests {
+			dec := yaml.NewYAMLOrJSONDecoder(strings.NewReader(manifest), 10)
+			obj := &unstructured.Unstructured{}
+			err := dec.Decode(obj)
+			if err != nil {
+				combination.SetStatusCondition(metav1.Condition{
+					Type:               v1alpha1.TypeInvalid,
+					Status:             metav1.ConditionFalse,
```
joelanford, line 178: This should be set to `True` (as in "this _is_ invalid"), right?
```suggestion
					Status:             metav1.ConditionTrue,
```

joelanford: Also, if a previous reconcile set this to `True`, but the user resolved the problem, when do we set it to false?

pkg/controller/combo_controller.go
```diff
@@ -153,6 +157,53 @@ func (c *combinationController) Reconcile(ctx context.Context, req ctrl.Request)
...
+					Reason:             v1alpha1.ReasonApplyFailed,
+					LastTransitionTime: metav1.NewTime(time.Now()),
+					Message:            fmt.Sprintf("failed to apply manifest %s to cluster: %s", obj.GetName(), err.Error()),
+				})
+				return reconcile.Result{}, errors.NewAggregate([]error{err, c.Status().Update(ctx, combination)})
+			}
+		}
+		// update status once all manifests have been applied
+		combination.SetStatusCondition(metav1.Condition{
+			Type:               v1alpha1.TypeFinished,
```
joelanford, line 199: If a previous reconcile finished successfully, but a subsequent reconcile fails, shouldn't we set this condition to `Finished` == `False`?

pkg/controller/combo_controller.go
tylerslaton: So this probably outlines a misunderstanding I had when writing this. Should every type of Condition be updated during reconciliation steps? Right now, this is just moving through the conditions and overwriting the current status on the list. Should it instead be having all of a set of relevant conditions precent (Invalid, Processing, Processed, etc.) and update the "Status" portion of that through the process?

joelanford: Lots of good detail here: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

To me, this is one of the harder things to get right. The general pattern I try to follow is:
1. Ensure every single condition supported by the operator is set before you return from reconcile.
2. Every single condition should be set based on the latest known state for that condition.
3. If you can't derive a definitive True/False the status of a particular condition, it should be set to "Unknown" with a description message about why it could not be determined (usually from an error that occurred somewhere)

exdx:
This will be overhauled to incorporate some changes to how combo updates the status of the combination CR

/hold

exdx:
Based on some offline conversations regarding this feature, it became not as evident as to why combo should apply manifests it generates on its own -- combo is then responsible for managing the lifecycle of arbitrary resources and dealing with all the nuances involved in that. Keeping combo small and simple in scope I believe is more preferable, and it should be up to clients of the API to handle the lifecycle of objects that are generated for their purposes. 

For now, I am going to close this PR, as there isn't enough desire to continue down this path. If we do feel that combo should in fact apply and manage arbitrary manifests then we can continue to work on this implementation.

Issue #79, opened 2022-02-17
panic in combo-operator when applying a combination with one value


exdx:
Steps to reproduce locally:
Apply template
```
apiVersion: combo.io/v1alpha1
kind: Template
metadata:
  name: cm-template
spec:
  body: |
    ---
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: cm
      namespace: default
    data:
      key: VALUE
```
Apply combination
```
apiVersion: combo.io/v1alpha1
kind: Combination
metadata:
  name: cm-feature
spec:
  template: cm-template
  arguments:
  - key: VALUE
    values:
    - "fox"
  apply: true
```

This causes a panic locally for the combo operator
```
panic: runtime error: index out of range [-1]

goroutine 199 [running]:
github.com/operator-framework/combo/pkg/combination.(*stream).nextIterativeCombination(0xc0006174a0)
	/Users/dsover/code/src/combo/pkg/combination/combination.go:113 +0x325
github.com/operator-framework/combo/pkg/combination.(*stream).Next(0xc000223c50, {0x0, 0x0})
	/Users/dsover/code/src/combo/pkg/combination/combination.go:212 +0x2e
github.com/operator-framework/combo/pkg/template.(*builder).Build(0xc0006cd980, {0x17ab318, 0xc000223c50})
	/Users/dsover/code/src/combo/pkg/template/builder.go:45 +0x7a
github.com/operator-framework/combo/pkg/controller.(*combinationController).Reconcile(0xc0000ab420, {0x17ab318, 0xc000223c50}, {{{0x0, 0x14b8ca0}, {0xc00041f278, 0xc0006e9f00}}})
	/Users/dsover/code/src/combo/pkg/controller/combo_controller.go:136 +0x478
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile(0xc00014c3c0, {0x17ab318, 0xc000223bc0}, {{{0x0, 0x14b8ca0}, {0xc00041f278, 0xc0006bb300}}})
	/Users/dsover/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.10.2/pkg/internal/controller/controller.go:114 +0x222
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc00014c3c0, {0x17ab270, 0xc000301f00}, {0x1457a60, 0xc0006eace0})
[...]

PR #81, opened 2022-02-22
fix: correct implementation of nextIterativeCombination method to handle singular input

### LINKED ISSUE #80: fix nextIterativeCombination method to correctly output combinations given a singular input
# Story
As a user of Combo
I need correct output of combinations given a singular combination input e.g `TEST1: "foo", "bar"` 
# Summary
We need to fix the bug in `nextIterativeCombination()` that is causing the output of the previous example to be `TEST1: "foo"` instead of the correct output `TEST1: "foo", TEST1: "bar"` This will ensure that `nextIterative` provides correct combinations. 
# Acceptance Criteria
* Correct output given the example test case
* Passes all tests

Note: This bug is only an issue in `nextIterativeCombination` and not `nextPreSolvedCombination`

jbuchananr:
This PR adds an edge case to `nextIterativeCombination()` method that correctly handles the case given an input of a single parameter

e.g
Input: `test: "foo", "bar"`
Output:
```---
test: foo
---
test: bar
```

closes #80

### DIFF
--- a/pkg/combination/combination.go
+++ b/pkg/combination/combination.go
@@ -86,6 +86,17 @@ func (cs *stream) nextIterativeCombination() (map[string]string, error) {
 	// comboList is a variable that holds a list of combinations to be returned.
 	comboList := map[string]string{}
 
+	// Edge case: 1 parameter
+	if len(cs.parameterListFromArgs) == 1 {
+		key := cs.parameterListFromArgs[0]
+		comboList[key] = cs.args[cs.parameterListFromArgs[0]][cs.positionsMapInArgs[0]]
+		cs.positionsMapInArgs[0]++
+		if cs.positionsMapInArgs[0] == len(cs.args[cs.parameterListFromArgs[0]]) {
+			cs.solved = true
+		}
+		return comboList, nil
+	}
+
 	// Generate the list of combinations based off current positions
 	for x := 0; x < len(cs.parameterListFromArgs); x++ {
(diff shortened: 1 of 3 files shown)


pkg/combination/combination.go
```diff
@@ -86,6 +86,17 @@ func (cs *stream) nextIterativeCombination() (map[string]string, error) {
 	// comboList is a variable that holds a list of combinations to be returned.
 	comboList := map[string]string{}
 
+	// Edge case: 1 parameter
```
tylerslaton, line 89: This will work but from a readability stand-point this function is getting a bit out of hand. I'd like to see some comments and have the logic broken up a bit where makes sense. For example, a couple of these for loops could potentially go into other functions or refactored.

tylerslaton (changes requested):
Good catch! I would like to see a unit test that checks for this new logic if you don't mind adding that.

pkg/combination/combination.go
jbuchananr: I think that's a great idea. Would we want the readability improved in this PR or in a separate issue?

tylerslaton: Let's make it a follow-up.

tylerslaton (approved):
Look great! Let's be sure to improve the readability of the `nextIterative` function in #82

exdx:
Looks like this closes #79?

Closes #79

PR #91, opened 2022-03-18
Version Command Does Not Display in Container

nsapse:
- Modified Dockerfile to include portions of .git directory necessary for
Git describe (the source of version) to successfully execute. Version command
was initially failing because git could not find the necessary information in
the .git directory at the root of the project.

- Added logging announcing version started as part of run command

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,11 @@ COPY pkg pkg
 COPY main.go main.go
 COPY Makefile Makefile
 
+# copy git-related information for binary version information
+COPY .git/refs .git/refs
+COPY .git/HEAD .git/HEAD
+COPY .git/objects .git/objects
+
 RUN make build-cli
 
 # Copy the binary over to a distroless image and run it
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"github.com/operator-framework/combo/pkg/controller"
+	"github.com/operator-framework/combo/pkg/version"
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -46,6 +47,7 @@ This will reconcile any events for the Combination and Template resources.
 			return err
 		}
 
+		rootLog.Info("Starting Combo", "combo version", version.ComboVersion, "git commit", version.GitCommit, "kubernetes version", version.KubernetesVersion)
 		return mgr.Start(signals.SetupSignalHandler())
 	},
 }
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,6 @@ GO_BUILD := $(Q)go build
 PKGS := $(shell go list ./...)
 COMBO_VERSION :=  $(shell git describe --match 'v[0-9]*' --tags --always)
 
-
 # Binary build options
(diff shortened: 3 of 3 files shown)


exdx (approved):
Would it be helpful to also emit a log on startup of what version combo is (can be done in a follow-up)?

cmd/run.go
```diff
@@ -46,6 +49,7 @@ This will reconcile any events for the Combination and Template resources.
 			return err
 		}
 
+		rootLog.Info(fmt.Sprintf("combo version %s started", version.ComboVersion))
```
timflannagan, line 52: ``
```suggestion
		rootLog.Info("starting combo", "version", version.ComboVersion)
```

Dockerfile
```diff
@@ -13,6 +13,11 @@ COPY pkg pkg
 COPY main.go main.go
 COPY Makefile Makefile
 
+# copy git-related information for binary version information
+COPY .git/refs .git/refs
+COPY .git/HEAD .git/HEAD
+copy .git/objects .git/objects
```
tylerslaton, line 19:
```suggestion
COPY .git/objects .git/objects
```

cmd/run.go
```diff
@@ -46,6 +47,7 @@ This will reconcile any events for the Combination and Template resources.
 			return err
 		}
 
+		rootLog.Info("starting combo", "version", version.ComboVersion)
```
tylerslaton, line 50: Should we also include the `KubernetesVersion` and `GitCommit`?

tylerslaton (changes requested):
Looks good! Just two small fixes to requestl.

cmd/run.go
timflannagan: It looks like there's already an existing String method that could dump out that information.

nsapse: What is that method, just out of curiosity?

timflannagan: https://github.com/operator-framework/combo/blob/main/pkg/version/version.go#L14-L18 is the method I had in mind. It just outputs a formatted string that contains the version information we pass when building the combo binary.

nsapse: I guess the logger takes in raw strings or doesn't process formatting characters for some other reason because when  passed in it results in a message with `\n` being printed explicitly instead of actual newlines (see here `{"level":"info","ts":1649706102.7005851,"msg":"Starting Combo","versions:":"Combo version: v0.0.1-4-g83a20a6\nGit commit: 83a20a646ce3f6562aa47ed90b16e3df81cbaf5e\nKubernetes version: v0.22.2"}`. Should I try and just manually input the version information for output?

timflannagan: Yep - let's just revert back to using the version.ComboVersion field directly.

cmd/run.go
```diff
@@ -46,6 +47,7 @@ This will reconcile any events for the Combination and Template resources.
 			return err
 		}
 
+		rootLog.Info("Starting Combo", "combo version:", version.ComboVersion, "git commit", version.GitCommit, "kubernetes version", version.KubernetesVersion)
```
tylerslaton, line 50: nit:
```suggestion
		rootLog.Info("Starting Combo", "combo version", version.ComboVersion, "git commit", version.GitCommit, "kubernetes version", version.KubernetesVersion)
```

tylerslaton (approved):

PR #92, opened 2022-03-29
Deflate the combo manifests directory

exdx:
This PR deflates the combo manifests directory (puts everything at the top-level) so that the combo operator bundle conforms to the plain bundle spec.


timflannagan (approved):
LGTM - let's get a second approval before merging.

PR #94, opened 2022-04-06
add: created flag to specify verbosity to combo CR controller

### LINKED ISSUE #62: add ability to specify log level for controller through command line
# Story
As a user of Combo
I would like to have a knob to adjust the verbosity of Combo
So that it is easy to filter how verbose its logs are

# Summary
We should expose a knob that enables changing how verbose Combo's CR controller is within its container. This will make working with the operator a better experience. This will most likely look like a command line flag of some sort.

# Acceptance Criteria
- [ ] Defined logging approach
- [ ] Clear logging levels
- [ ] Knob for changing the verbosity

jbuchananr:
## Story
Users are not able to adjust the verbose of Combo's CR controller. A way to adjust this verbosity needs to be implemented in order to allow users to define their logging. 

## Summary
* add flag in the `run` command that allows user to specify verbosity level by calling e.g. `... run --loglvl 3`

## Acceptance Criteria 
* Defined logging approach
* Clear logging levels
* Knob for changing the verbosity

Closes #62

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -8,6 +8,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 )
 
+func init() {
+	runCmd.Flags().Int("verbosity", 1, "Sets verbosity level of combo CR controller with default verbosity 1. Verbosity decreases as the value given increases.")
+}
+
 var runCmd = &cobra.Command{
 	Use:   "run",
 	Short: "Run Combo as a controller on the cluster",
@@ -25,9 +29,14 @@ This will reconcile any events for the Combination and Template resources.
 			return err
 		}
 
+		verbosityLevel, err := cmd.Flags().GetInt("verbosity")
+		if err != nil {
+			return err
+		}
+
 		c, err := controller.NewController(
 			mgr.GetClient(),
-			ctrl.Log.WithName("run"),
+			ctrl.Log.V(verbosityLevel).WithName("run"),
 		)
 		if err != nil {
 			return nil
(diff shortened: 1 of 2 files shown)


cmd/run.go
```diff
@@ -8,6 +8,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 )
 
+func init() {
+	evalCmd.Flags().Int("loglvl", 1, "Sets verbosity level of combo CR controller")
```
tylerslaton, line 12: Could we name this log `verbosity` instead of `loglvl`? I feel like it is a bit more indicative of what we're setting with this. Also, it may make sense in the description to add that the higher the number, the less verbose it is.

tylerslaton (commented):
Looks good! One super small comment. Looks like the E2E suite is failing as this change is preventing the Combo controller from starting.

cmd/run.go
```diff
@@ -25,9 +29,14 @@ This will reconcile any events for the Combination and Template resources.
...
 
+		verbosityLevel, err := cmd.Flags().GetInt("loglvl")
+		if err != nil {
+			return err
+		}
+
 		c, err := controller.NewController(
 			mgr.GetClient(),
-			ctrl.Log.WithName("run"),
+			ctrl.Log.V(verbosityLevel),
```
tylerslaton, line 39:
```suggestion
			ctrl.Log.V(verbosityLevel).WithName("run"),
```

jbuchananr: Got it.

cmd/run.go
jbuchananr: Sounds good to me.

tylerslaton (approved):
Nice! Looks great. This will really help give users the ability to tweak the log level 👍🏻

Issue #96, opened 2022-04-14
Automate building and pushing images to quay.io


oceanc80:
For ease of keeping images in quay.io up to date, it would be helpful to automate building and pushing images of any tagged versions of Combo to quay.io.

Issue #84, opened 2022-03-01
add support for plain bundle format


tylerslaton:
Since RukPak is starting to come to its own, it makes sense for Combo to be a first example bundle for its plain manifests bundle format. Proof of concept for this has already been done via the `quay.io/tyslaton/combo-plain-bundle:v0.0.1` image.

It was constructed using the following Dockerfile:

```dockerfile
FROM scratch
LABEL manifests="/manifests"
COPY ./manifests /manifests
```

Which loads in the manifests from `combos` manifest directory. As a result, we would like an official support for this so that RukPak can use it in its internal documentation as an example for users to get started with.

# Acceptance Criteria
- [x] Release schedule determined
- [x] Dockerfile.plainbundle
- [x] Make target
- [x] Necassary CI actions

exdx:
There is a gap in documentation in what a plain/k8s bundle looks like right now, and having combo as an example is very helpful in the effort to document this new format. I do think we need to settle on a spec, for example the manifests location and the label name,  as a prereq for really publishing this new format. combo should be an example of the spec, and since it's the very first officially supported plain bundle, it's important to have it match exactly what we intend with the new bundle format.

exdx:
Update: the plain bundle spec landed in https://github.com/operator-framework/rukpak/pull/149 and the combo manifests directory was updated to conform to the spec in #92. At this point combo is equipped as a plain bundle, but further actions defined in the A/C are still todos.

oceanc80:
For now, releases will be triggered manually via versioned releases and every merge to main. This issue tracks the need for a regular release cadence going forward as Combo matures: https://github.com/operator-framework/combo/issues/100

PR #97, opened 2022-04-28
Add automated releasing

oceanc80:
Add automated build/release on updates to the main branch and any new tags

Closes https://github.com/operator-framework/combo/issues/96


.github/workflows/release.yaml
```diff
@@ -0,0 +1,57 @@
...
+          echo GORELEASER_ARGS="--rm-dist --skip-validate" >> $GITHUB_ENV
+        elif [[ $GITHUB_REF == refs/pull/* ]]; then
+          # PR build.
+          echo IMAGE_TAG="pr-$(echo "${GITHUB_REF}" | sed -E 's|refs/pull/([^/]+)/?.*|\1|')" >> $GITHUB_ENV
+        else
+          echo IMAGE_TAG="$(git describe --tags --always)" >> $GITHUB_ENV
+        fi
+
+    - name: Run goreleaser
+      run: make release
```
timflannagan, line 55: I think we'd need to add this release target to the root Makefile, or use the first-class goreleaser github action.

.goreleaser.yaml
```diff
@@ -0,0 +1,73 @@
...
+  - "{{ .Env.IMAGE_REPO }}:{{ .Env.IMAGE_TAG }}-s390x"
+checksum:
+  name_template: 'checksums.txt'
+snapshot:
+  name_template: "{{ incpatch .Version }}-next"
+changelog:
+  use: github-native
+  skip: $DISABLE_RELEASE_PIPELINE
+release:
+  disable: $DISABLE_RELEASE_PIPELINE
```
timflannagan, line 73: Running goreleaser directly using this .goreleaser.yaml configuration would be problematic as there's no layer that sits between evaluating the value of this variables, and the source of truth configuration file that goreleaser relies on when running it's tooling. We can either pivot to the rukpak implementation that's centered around checking a goreleaser template file, and using envsubst to evaluate these values at runtime and produces a static .goreleaser.yaml configuration file, or disable the release/changelog pipelines explicitly.

The latter implementation would require further updates to the workflow such that automation creates a release for us during tag creation triggers.

.goreleaser.yaml
```diff
@@ -0,0 +1,73 @@
+env:
+- GOPROXY=https://proxy.golang.org|direct
+- GO111MODULE=on
+- CGO_ENABLED=0
+- PKG=github.com/operator-framework/combo/pkg/version
+- KUBERNETES_VERSION=v0.22.2
```
timflannagan, line 6: This feels potentially brittle. Can we try and use the Makefile as the source of truth here, and export any variable that sets the Kubernetes version (if that variable even exists today)?

timflannagan:
I had one more comment but I think we're getting close.

.goreleaser.yaml
timflannagan: Yup that's more-or-less what I had in mind.

oceanc80: I think so. I think one of the other variables is exported that way already.

Makefile
```diff
@@ -110,3 +110,15 @@ run-local: build-local-container load-image deploy ## Run Combo on local environ
...
+##################
+# Release Targets#
+##################
+export DISABLE_RELEASE_PIPELINE ?= true
+substitute:
+	envsubst < .goreleaser.template.yml > .goreleaser.yml
+
+release: GORELEASER ?= goreleaser
+release: GORELEASER_ARGS ?= --snapshot --rm-dist
+release: goreleaser substitute
```
timflannagan, line 123: I think we'd need to remove the `goreleaser` target as a prerequisite target as that target doesn't appear to exist?

oceanc80: Oops... missed that in the copy/paste clean up.

timflannagan (approved):

tylerslaton (approved):
Nice! This looks great. I think the only thing we need now is credentials. @joelanford, would you mind setting those up before we merge?

tylerslaton:
Just thought of a follow-up for this. We'll probably want to create a Bundle format that we can submit for both OLM and RukPak. Not required for this PR but I'll create a follow-up.

oceanc80:
> Just thought of a follow-up for this. We'll probably want to create a Bundle format that we can submit for both OLM and RukPak. Not required for this PR but I'll create a follow-up.

@tylerslaton Does [this issue](https://github.com/operator-framework/combo/issues/84) cover that?

tylerslaton:
@oceanc80 It sure does, good catch!

PR #99, opened 2022-05-05
Automate bundle build and push

oceanc80:
This automates building and pushing the Combo bundle when a new version is tagged.

Closes https://github.com/operator-framework/combo/issues/84


tylerslaton (commented):
Just a few comments I wanted your opinion on. Other than that, this is basically ready to go!

.github/workflows/bundle.yaml
```diff
@@ -0,0 +1,35 @@
+name: bundle
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  bundle:
+    name: bundle
```
exdx, line 10: nit: I believe the name of this job would be `bundle/bundle` in the actions UI -- do we want to name it something more descriptive? Maybe `bundle/deploy` is appropriate?

.github/workflows/bundle.yaml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Docker Login
+      if: ${{ github.event_name != 'pull_request' }}
```
exdx, line 14: nit: do we need this `if` check if the workflow is only triggered on new tag pushes defined above?

Dockerfile.plainbundle
```diff
@@ -0,0 +1,3 @@
+FROM scratch
+LABEL manifests="/manifests"
```
exdx, line 2: nit: I would avoid adding labels until we have some kind of consensus around metadata and whether the labels are helpful. Right now we have the limitation that for a given plain+v0 bundle image to be valid it must have a `/manifests` directory at the root, so this label is redundant.

Dockerfile.plainbundle
```diff
@@ -0,0 +1,3 @@
...
+COPY ./manifests /manifests
```
exdx, line 3: nit: newline

.github/workflows/bundle.yaml
```diff
@@ -0,0 +1,35 @@
...
+      uses: docker/login-action@v1
+      with:
+        registry: quay.io
+        username: ${{ secrets.QUAY_USERNAME }}
+        password: ${{ secrets.QUAY_PASSWORD }}
+
+    - name: Set the release related variables
+      run: |
+        if [[ $GITHUB_REF == refs/tags/* ]]; then
```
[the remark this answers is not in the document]
oceanc80: Did we want to push the bundle on updates to main as well as version tags or just on version tags? I think @exdx also had some thoughts here.

Makefile
```diff
@@ -91,12 +98,15 @@ test-e2e: ## Run the e2e tests
...
 # Running Targets #
 ###################
-.PHONY: load-image deploy teardown run run-local run-e2e run-e2e-local
+.PHONY: load-image load-bundle deploy teardown run run-local run-e2e run-e2e-local
 
 IMAGE_LOAD_COMMAND=kind load docker-image
 load-image: ## Load-image loads the currently constructed image onto the cluster
 	$(IMAGE_LOAD_COMMAND) $(IMAGE)
 
+load-bundle: ## Loads the currently constructed Combo bundle onto the cluster
```
exdx, line 107: hmm, this target is very similar to the one above, I wonder if it adds enough value to include it? Or we can simply advise to `make load-image IMAGE=<bundle-image>` instead?

exdx (approved):
Nice!

Dockerfile.plainbundle
```diff
@@ -0,0 +1,3 @@
+FROM scratch
```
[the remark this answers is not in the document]
oceanc80: We could probably combine the .goreleaser and .local ones as they're pretty similar.

.github/workflows/bundle.yaml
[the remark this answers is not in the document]
exdx: I think we should push the bundle only on tagged releases like the way it's implemented currently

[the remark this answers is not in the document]
tylerslaton: Ah, right that makes sense as these manifests will stay the same most of the time. Will resolve this comment!

Dockerfile.plainbundle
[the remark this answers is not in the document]
tylerslaton: Maybe we could do that and also make local way of building the controller container be the default like RukPak. That would consolidate three of them together. Thoughts?

Makefile
oceanc80: Yeah I think as long as we document it well that's totally reasonable.

Dockerfile.plainbundle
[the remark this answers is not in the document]
oceanc80: Yeah, that seems reasonable to me.

tylerslaton (approved):
Nice! This a huge improvement to the maintainability of this repo. Great contribution 🙏🏻

Makefile
```diff
@@ -64,15 +68,15 @@ verify: tidy generate format lint ## Verify the current code generation and lint
 
 VERSION_FLAGS=-ldflags "-X $(VERSION_PATH).GitCommit=$(GIT_COMMIT) -X $(VERSION_PATH).ComboVersion=$(COMBO_VERSION) -X $(VERSION_PATH).KubernetesVersion=$(KUBERNETES_VERSION)"
 build-cli: ## Build the CLI binary. Specify VERSION_PATH, GIT_COMMIT, or KUBERNETES_VERSION to change the binary version. You may also specify BUILD_OS and BUILD_ARCH to change the build's binary. 
-	$(Q) CGO_ENABLED=0 GOOS=$(BUILD_OS) GOARCH=$(BUILD_ARCH) go build $(VERSION_FLAGS) -o ./bin/combo
+	$(Q) CGO_ENABLED=0 GOOS=$(BUILD_OS) GOARCH=$(BUILD_ARCH) go build $(VERSION_FLAGS) -o ./combo
```
exdx, line 71: nit: I think the convention is to put the binary under `/bin` -- is there a reason to put it at the top level?

exdx (approved):
Nice job consolidating the dockerfiles!

Makefile
oceanc80: I ran into some issues with the different binary locations between local/CI builds and ended up copying what is in RukPak. The final container has the binary under `/bin`.

PR #101, opened 2022-05-11
.github,Makefile: Fix the bundle workflow

timflannagan:
Update the root Makefile and ensure the image-related variables can be overridden. Overhaul the bundle.yaml GHA workflow to fix failures that arose in https://github.com/operator-framework/combo/actions/runs/2309116312.

### DIFF
--- a/.github/workflows/bundle.yaml
+++ b/.github/workflows/bundle.yaml
@@ -10,6 +10,9 @@ jobs:
     name: release
     runs-on: ubuntu-latest
     steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
     - name: Docker Login
       uses: docker/login-action@v1
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/bundle.yaml
```diff
@@ -10,6 +10,9 @@ jobs:
     name: release
     runs-on: ubuntu-latest
     steps:
+    - name: Checkout
+      uses: actions/checkout@v2
```
timflannagan, line 14: We ran into issues where the docker/build-push-action wasn't able to locate the Dockerfile.plainbundle, and adding the checkout step before running that job silenced that error.

.github/workflows/bundle.yaml
```diff
@@ -25,11 +28,13 @@ jobs:
         else
           echo BUNDLE_TAG="$(git describe --tags --always)" >> $GITHUB_ENV
         fi
-        echo BUNDLE="quay.io/operator-framework/combo-bundle:${BUNDLE_TAG}" >> $GITHUB_ENV
+        echo BUNDLE_REPO="quay.io/operator-framework/combo-bundle" >> $GITHUB_ENV
```
timflannagan, line 31: When printing out the $BUNDLE variable in a subsequent step, the tag value was empty, so we changed this variable to reflect the repository name, and concatenated the repo:tag value in the docker/build-push-action step.

Makefile
```diff
@@ -14,13 +14,13 @@ COMBO_VERSION :=  $(shell git describe --match 'v[0-9]*' --tags --always)
 export KUBERNETES_VERSION=v0.22.2
 
 # Container build options
-export IMAGE_REPO=quay.io/operator-framework/combo-operator
-export IMAGE_TAG=latest
-IMAGE=$(IMAGE_REPO):$(IMAGE_TAG)
+export IMAGE_REPO ?= quay.io/operator-framework/combo-operator
+export IMAGE_TAG ?= latest
```
timflannagan, line 18: This was necessarily as the release GHA workflow uses the `release` Makefile target, and we were always using the latest value despite overriding this value during that workflow.

oceanc80 (approved):

